PR #44, opened 2017-04-05
BCN-ISAK-ENGDAHL

isaklafleur:


starter-code/css/main.css
```diff
@@ -6,3 +6,156 @@ Green: #00B172
...
+
+
+.backgroundimage {
+    background-image: url("../images/landing.jpg");
+    background-size: contain;
+    height: 350px;
+    text-align: center;
+    display: flex;
+    align-items: center;
+}
```
tawebbcn: Check the background-repeat: no-repeat; property, so the image won't be repeated

starter-code/css/main.css
```diff
@@ -6,3 +6,156 @@ Green: #00B172
...
+    vertical-align: middle;
+}
+
+
+.links .left-side h1 {
+      display: inline-block;
+      border-bottom: 5px solid white;
+}
+
+
```
tawebbcn: Good job!

PR #51, opened 2017-04-06
BCN-Giuseppe-Fierro

gfrr:


starter-code/index.html
```diff
@@ -2,41 +2,103 @@
...
+      <div id="right">
+        <img src="images/spotify-app.jpg">
+
+      </div>
+    </div>
+  </div>
+  </footer>
+
 </body>
 </html>
```
tawebbcn: #checked. Nicely done.

PR #85, opened 2017-06-14
BCN: Miguel

Mi6u3l:


tawebbcn:
Navigation should be fixed at the top.
Try use more semantic tags if it's possible.
Nice logo animation 👍 
The What's on Spotify & Kanye West section are not centered. just align to the left.
You missed the white spotify logo in the Kanye West section
Nice CSS keeping 2 level of specification 👍 
Hexadecimal values like `#ffffff , #aabbcc` can be shorted as `#fff , #abc`

PR #98, opened 2017-06-14
bcn-maria-mayer

mariamayer:


tawebbcn:
Try to leave an empty line after each css rule, for more readability.
See 2 weirds div.row  but they are empty, shouldn't the .box elements be his children ?
The navbar should be fixed at the top.
nice css 👍 
semantic tags 👍

PR #110, opened 2017-06-15
BCN-julius-suominen

jsur:
done


matilore:
well done, i like the bouncing image, good

PR #111, opened 2017-06-15
BCN Stephanos Spotify Clone

stephanos7:


tawebbcn:
**HTML** 
Your `landing-container `could be the header right ? instead of a div
You have a comment like: 
```
<!-- SECONDARY SECTION -->
<!-- Green Section -->
```
if these blocks of code are sections for you, why didn't use `section` tag instead of `div`
Don't leave spaces like `class = "`, right way: `class="`  
Alt attributes filled 👍 
Good indentation 👍 

**CSS**
Close the css rule`  } ` right after the last property, without empty lines.
Empty line between css rules 👍 
Try to avoid selecting tag element in the first level, use classes instead.
zero values without unit 👍 

Good job

matilore:
In order to make the image bouncing use animations and key frames, applying a margin to an hover selector is not a proper solution

PR #113, opened 2017-06-15
bcn-denis-barriere

denisbarriere:
done


tawebbcn:
Well done 👍 
usually in html we use the dash syntax, and camel case for JS, but it's really up to you 😃

denisbarriere:
I need to read about BEM naming convention, but I could not find the time so far!!

PR #115, opened 2017-06-15
BCN- Alex Fontanellas

alexfontanellas:


tawebbcn:
Try avoid select html tags in the CSS in the first level, use classes instead.
Try use semantics html tags when it's possible instead of only divs, like main, section, article...
Nice logo animation 👍 
Most of the elements are not properly centered.

PR #116, opened 2017-06-15
BCN - Arnau_Guadall

arnauguadall:


tawebbcn:
You could `.footer-left ` use this class to apply  the css properties to all his children with the class `.margin-left`  like:
```
.footer-left h3,
.footer-left p {
 // Margin left properties here
}
```
Avoiding all that .margin-left class repetitions
Don't use the <center> for style the text, simply use css, maybe in the future you gonna want that block not in the center, so just change a css property instead of deleting a html tag by another one.

Good css 👍

PR #138, opened 2017-08-16
BCN - Philip Wisner

philipwisner:
It should be all finished. The only issues is the spotify logo on the bottom green part messes with the responsive design. Hector said not to worry about it.


tawebbcn:
Good job! :+1: 

Today you will learn how to do it total responsive.

PR #156, opened 2017-09-06
PAR - Jad Challita - Iteration 1, done!

jadch:


starter-code/css/main.css
```diff
@@ -5,4 +5,176 @@ Text: 1A1A1A
 Green: #00B172
 White: #FFF
 
+font: sans-serif
 */
+
+* body {
```
yacinehmito: I don't think you need the `*` here.

PR #157, opened 2017-09-06
PAR stuart-gardner

stuartogardner:


starter-code/css/main.css
```diff
@@ -6,3 +6,175 @@ Green: #00B172
...
+  color: #FFF;
+  margin-top: 50px;
+}
+
+.main-image h1{
+  text-align: center;
+  font-size: 5em;
+  position: relative;
+  padding-top: 21%;
+  padding-bottom: -10%;
```
yacinehmito: There are no negative paddings. There are negative margins though.

starter-code/css/main.css
```diff
@@ -6,3 +6,175 @@ Green: #00B172
...
+}
+
+.nav-list{
+  margin-right: 5%;
+}
+
+.main-image{
+  background-image: url("../images/landing.jpg");
+  background-repeat: no-repeat;
+  background-size: contain;
```
yacinehmito: Try `background-size: cover` with `background-position: center` and see if you like that better when resizing the window.

PR #158, opened 2017-09-06
PAR - Thibaut Davoult - Iteration1

thibautdavoult:
green area is better than nothing but the code aint pretty.

### DIFF
--- a/starter-code/css/main.css
+++ b/starter-code/css/main.css
@@ -105,33 +105,35 @@ body {
 
 .yeezy-green {
   background-color: #00B172;
-  /*display: flex;
-  justify-content: space-between;*/
-  margin: 0 30px 30px 30px;
+  margin: 30px;
   position: relative;
-  width: 94%;
 }
 
-/*
+.yeezy-phone {
+  position: absolute;
+  /*margin-top: 100px;*/
+  right: 80px;
+  top: 80px;
 
-issues on logo and image
+}
 
-*/
+.yeezy-phone img {
+  height: 500px;
+}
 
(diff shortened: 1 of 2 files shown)


starter-code/index.html
```diff
@@ -2,41 +2,78 @@
...
+    <ul class="nav-ul">
+      <li>Premium</li>
+      <li>Discover</li>
+      <li>Help</li>
+      <li>Download</li>
+    </ul>
+  <nav>
+  </div>
+  <div class="clearfloats"></div>
+  <content>
```
yacinehmito: Don't use this. It doesn't mean what you think it does. `div` is decent enough here. `section` or `main` are semantic alternatives.

starter-code/css/main.css
```diff
@@ -6,3 +6,157 @@ Green: #00B172
...
+  top: 80px;
+
+}
+
+.yeezy-phone img {
+  height: 500px;
+}
+
+.yeezy-logo {
+  position: absolute;
```
yacinehmito: *This* is a decent use of `position: absolute`.

starter-code/css/main.css
```diff
@@ -6,3 +6,157 @@ Green: #00B172
...
+}
+
+.yeezy-green {
+  background-color: #00B172;
+  margin: 30px;
+  position: relative;
+}
+
+.yeezy-phone {
+  position: absolute;
```
yacinehmito: Don't use `position: absolute`if you can avoid it. Here, flexbox or floats can do the job.

PR #160, opened 2017-09-06
PAR - Liudmyla - Iefremova

LiudmilaE:


starter-code/css/main.css
```diff
@@ -6,3 +6,190 @@ Green: #00B172
 White: #FFF
 
 */
+
+*{
+  margin: 0;
+  padding: 0;
+  color: 1A1A1A;
```
yacinehmito: You really need the `#`.

starter-code/css/main.css
```diff
@@ -6,3 +6,190 @@ Green: #00B172
 White: #FFF
 
 */
+
+*{
```
yacinehmito: `body`would be a better fit that `*`.

PR #163, opened 2017-09-07
PAR OMBLINE GILLIOT first sub

ombline:


starter-code/css/main.css
```diff
@@ -6,3 +6,112 @@ Green: #00B172
...
+
+header {
+  padding: 10px 40px 0px 40px;
+}
+
+header img {
+  width: 14%;
+}
+
+ul {
```
yacinehmito: A more specific selector would be more appropriate. You don't want those rules to affect `every` list on the page (even though there is only one yet).

starter-code/css/main.css
```diff
@@ -6,3 +6,112 @@ Green: #00B172
...
+  margin: 0;
+}
+
+/*HEADER SECTION*/
+
+header {
+  padding: 10px 40px 0px 40px;
+}
+
+header img {
```
yacinehmito: Do you really want the logo to shrink or expand when the user resizes the window?

starter-code/css/main.css
```diff
@@ -6,3 +6,112 @@ Green: #00B172
...
+}
+
+/*SECTION ONE*/
+
+.music {
+position: relative;
+}
+
+.texte {
+position: absolute;
```
yacinehmito: It works really well in your case but I'm sure there are simpler way to do what you want without a `position: absolute`.

starter-code/css/main.css
```diff
@@ -6,3 +6,112 @@ Green: #00B172
...
+}
+
+/* SECTION 2*/
+.what {
+  text-align: center;
+}
+
+.title h1 {
+  font-size: 35px;
+  text-decoration: underline;
```
yacinehmito: Yes. However to match the screenshot perfectly a border would be more appropriate.

starter-code/css/main.css
```diff
@@ -115,3 +124,52 @@ width: 1000px;
...
+/*SECTION THREE*/
+
+.yeezy {
+  margin: 0 50px 80px 50px;;
+  background-color: #00B172;
+  display: flex;
+}
+
+.yeezy img {
+  width: 50%;
```
yacinehmito: Try to resize the window and you'll see what happens ;)

starter-code/css/main.css
```diff
@@ -115,3 +124,52 @@ width: 1000px;
...
+  margin: 100px 50px 50px 200px;
+}
+
+.yeezy h2, h3, p {
+  color: white;
+}
+
+.yeezy h2 {
+  font-size: 35px;
+  text-decoration: underline;
```
yacinehmito: `text-decoration: underline;`is a great idea! In our case a border might be more appropriate though.

PR #164, opened 2017-09-07
PAR - Frederic Magnin

frederic117:
:)

### DIFF
--- a/starter-code/css/main.css
+++ b/starter-code/css/main.css
@@ -6,3 +6,140 @@ Green: #00B172
 White: #FFF
 
 */
+
+nav{
+  top: 0px;
+  background-color: white;
+  position: fixed;
+  display: flex;
+  height: 80px;
+  padding-bottom: 50px;
+}
+
+.logo img{
+  max-width:30%;
+}
+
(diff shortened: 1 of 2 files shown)


starter-code/css/main.css
```diff
@@ -6,3 +6,140 @@ Green: #00B172
...
+  top: 0px;
+  background-color: white;
+  position: fixed;
+  display: flex;
+  height: 80px;
+  padding-bottom: 50px;
+}
+
+.logo img{
+  max-width:30%;
```
yacinehmito: Do you want the logo to change size when the user resizes the window?

starter-code/css/main.css
```diff
@@ -6,3 +6,140 @@ Green: #00B172
...
+  animation: bounce 1s;
+  animation-iteration-count: infinite;
+  animation-direction: alternate;
+  }
+
+.navText{
+  align-items: center;
+}
+
+ul{
```
yacinehmito: This selector works but wouldn't be specific enough in a real project. Use a class or a more specific selector so that your CSS rules don't affect other `ul`s.

PR #222, opened 2017-10-19
MIA Alejandra Del Valle Spotify Clone

alexdelvalle:
I'll be adding the bonus section tonight in a second update.


dsoraluz:
Great job with this exercise. Your attention to detail and being able to replicate what you see with css is apparent and it really shows with this one. Keep it up!

PR #227, opened 2017-10-20
MIA Alejandro Roman

aromanarguello:


dsoraluz:
Good job with this exercise. Don't worry about not getting the positioning just right yet. We will learn about more tools that you can use that will make positioning easier.

PR #228, opened 2017-10-20
Marcos Palacios: my spotify page

ShinCero:


dsoraluz:
Positioning in css is a tricky thing but with time and practice you will master it. We are going to be learning about some tools that will make positioning a lot easier so dont worry.

PR #229, opened 2017-10-20
MIA-Joshua_Jamison 

jamison05:
almost finished


dsoraluz:
Getting things to look the way we want with HTML and CSS is tough so don't worry, you will get it with some time and practice. We are going to be learning some things down the line that are going to make the process much easier so don't worry.

jamison05:
Thanks, I will work on the weekend in this.

PR #231, opened 2017-10-20
MIA Chu

Chu13:


dsoraluz:
Great job with this exercise. It looks almost identical to the orignal and your attention to detail and ability to do it in css shows. Keep up the good work!

PR #232, opened 2017-10-20
MIA Armando Acosta

arman66:


dsoraluz:
Positioning in HTML and CSS is tough but you will master it with some time and practice. We will be learining about some really helpful tools that will make it all easy so don't worry.

PR #245, opened 2017-10-27
@maxence @lucamerzi

lucamerzi:


starter-code/css/main.css
```diff
@@ -1,8 +1,112 @@
...
+li{
+	list-style: none;
+	display: inline;
+    margin-left: 10px;
+}
+
+nav{
+	height: 100px;
+	margin-bottom: 40px;
+/*	position: static;
```
yacinehmito: You want `positon: fixed;` here. Then, you need to position it using `top: 0`.

The problem then will be that the picture below will go underneath the navbar. Indeed, because `nav` has now it fixed position, it is not affecting the normal flow of the page. The next child, which is the `div` with the picture, will then be anchored to the top of the document. We need to move the rest of the page down to an amount equal to the height of the `nav` (here `100px`) by using `margin-top` on it.

Lastly, to make sure that the `nav` is always visible, we need to increase its `z-index`. It is zero by default. To set it to 1, we can add `z-index: 1;`.

starter-code/css/main.css
```diff
@@ -1,8 +1,112 @@
...
+	background-attachment: unset;
+    background-position: center;
+
+/*	background-color: blue;
+*/	height: 700px;
+}
+
+.picture h1, .picture p{
+	color: white;
+    width: 400px;
```
yacinehmito: There is no need to set a fixed width to those elements. You can remove `width: 400px;`, `margin: 10px auto;` and `text-align: center;`. Then, to center, add `text-align: center;` to the *parent*, which is `.picture`.

starter-code/css/main.css
```diff
@@ -1,8 +1,112 @@
...
+    background-position: center;
+    background-size: 10%;
+}
+
+.kanye-pic{
+	width: 80%;
+}
+
+.three-logos{
+	width: 40%;
```
yacinehmito: You can also set a fixed height to better align the text below.

starter-code/css/main.css
```diff
@@ -1,8 +1,112 @@
...
+	clear: both;
+	background-image: url("../images/spotify-icon-white.png");
+	background-repeat: no-repeat;
+    background-attachment: unset;
+    background-position: center;
+    background-size: 10%;
+}
+
+.kanye-pic{
+	width: 80%;
```
yacinehmito: Very good. Also, do you know the CSS properties `min-width` and `max-width`? With them you can make sure that the picture is not too small on small screens and not too big on big screens.

PR #246, opened 2017-10-27
PAR - @Medhi @maxence

Medhi-PLM:


starter-code/css/main.css
```diff
@@ -6,3 +6,189 @@ Green: #00B172
 White: #FFF
 
 */
+@import url('https://fonts.googleapis.com/css?family=Open+Sans');
+
+*{
```
yacinehmito: This will be applied to *every* object. I suggest you only apply it to `body`. Two reasons:

- Most elements have default margins and paddings set to 0. Only `body` has some default margin.
- If you set the `font-family` for an element, the effect is going to be propagated to the children. By setting it for `body` you will affect the whole page.

starter-code/css/main.css
```diff
@@ -6,3 +6,189 @@ Green: #00B172
...
+  background-repeat: no-repeat;
+  background-size: cover;
+  height: 750px;
+  position: relative;
+}
+
+h1 {
+  font-size: 4em;
+  color: white;
+  padding: 250px 500px 80px;
```
yacinehmito: You are centering the title manually. It may look great on your computer but it's not guaranteed to be the case for others.

yacinehmito: You are centering the title manually. It may look great on your computer but it's not guaranteed to be the case for others.

starter-code/index.html
```diff
@@ -3,40 +3,93 @@
...
-Music for everyone.
-Spotify is now free on mobile, tablet and computer.
-Listen to the right music, wherever you are.
+    <ul>
+      <li>Premium</li>
+      <li>Discover</li>
+      <li>Help</li>
+      <li>Download</li>
+    </ul>
+  </nav>
```
yacinehmito: Your `nav` is missing a `div` with the class `clearfix`.

starter-code/css/main.css
```diff
@@ -6,3 +6,189 @@ Green: #00B172
...
+@import url('https://fonts.googleapis.com/css?family=Open+Sans');
+
+*{
+  margin: 0;
+  padding: 0;
+  font-family: 'Open Sans';
+}
+
+nav {
+  position: fixed;
```
yacinehmito: When you use `position: fixed;` or `position: absolute;` you are excepted to manually position the element. Here, because you want the `nav` to stick at the top of the screen, you should add `top: 0;`.

Another problem is that your `nav` may then disappear behind the image below. To make sure it is always visible, you should give a big enough value to the `z-index` CSS property. Here, `z-index: 1;` would work.

PR #273, opened 2018-01-10
PAR Rithy spotify v1

Bentchaw:


ta-web-paris:
Some part good be improved but globally you mangage to display everything. Good job!

Maxence

PR #274, opened 2018-01-10
PAR-WD-Manuel-Rojas

manuelrs:


ta-web-paris:
Good job Manuel, you mastered the exercise :)

One tip I can give you is that you should "reset" your CSS. You can for example add the following code to remove the white space all around your page:
```
body {
    margin: 0px;
}
```

Maxence

PR #282, opened 2018-01-10
BCN - Josep Salvat

Jsalvat:


starter-code/css/main.css
```diff
@@ -6,3 +6,135 @@ Green: #00B172
...
+.mid-container{
+    width: 1200px;
+    height: 600px;
+    background-image: url("../images/landing.jpg");
+    background-size: 1200px 600px;
+}
+.first-img{
+    width: 250px;
+    height: 400px;
+    background-image: url("../images/music-icon.png");
```
tawebbcn: Not best practise to set icons as background images

starter-code/css/main.css
```diff
@@ -6,3 +6,135 @@ Green: #00B172
 White: #FFF
 
 */
+html {
+    box-sizing: border-box;
+  }
+  *, *:before, *:after {
+    box-sizing: inherit;
+  }
+
```
tawebbcn: If you inspect your body tag, you'll notice that it has an ugly thin margin. Try to remove it by setting margin and padding 0

starter-code/index.html
```diff
@@ -3,40 +3,64 @@
 <head>
   <meta charset="UTF-8">
   <title>Spotify</title>
+  <link rel="stylesheet" type="text/css" href="./css/main.css">
 </head>
 <body>
+<div class="wrapper">
+  <div class="header">
```
tawebbcn: maybe you need to put this outside the wrapper div

starter-code/index.html
```diff
@@ -3,40 +3,64 @@
...
+    <div class="logo-div"></div>
+  </div>
+  <div class="mid-container"></div>
+  <div class="imgs">
+    <p class="titleSpot">What’s on Spotify?</p>
+    <div class="first-img">
```
tawebbcn: instead of setting one class for each elements, you can set just one class for all of three divs, then you can style them reusing code!

PR #286, opened 2018-01-10
BCN - Jana

jjanastinn:


starter-code/css/main.css
```diff
@@ -6,3 +6,237 @@ Green: #00B172
...
+    box-sizing: border-box;
+  }
+  *, *:before, *:after {
+    box-sizing: inherit;
+  }
+
+body {
+    margin: 0;
+    padding: 0;
+    width: 100%;
```
tawebbcn: Don't touch body size. By default it will fill 100vw

starter-code/index.html
```diff
@@ -3,40 +3,66 @@
...
+      <div class="teaser-heading">
+        <h1>Music for everyone.</h1>
+        <p>Spotify is now free on mobile, tablet and computer.<br> Listen to the right music, wherever you are.</p>
+      </div>
+    </div>
+    <div class="main-content">
+      <h2>What’s on Spotify?</h2>
+      <hr>
+      <div class="content-info">
+        <div class="content-music">
```
tawebbcn: You can reuse styling code here! Better set the same class for all the divs that look the same.

PR #287, opened 2018-01-10
PAR Josh 

joshvardey:


ta-web-paris:
The beginning of the exercise is good. Keep working, you are on the good way :)

Maxence

PR #290, opened 2018-01-10
BCN - Pere

pereamoros:


starter-code/css/main.css
```diff
@@ -1,8 +1,134 @@
...
+    width: 350px;
+}
+.green-title {
+    font-size: 26px;
+    padding-bottom: 10px;
+    margin-bottom: 50px;
+    display: inline-block;
+    border-bottom: 5px solid #fff;
+}
+.green-subtitle {
```
tawebbcn: Try to organize your css structure like reset, typography, layout, components

PR #293, opened 2018-01-10
BCN - LUIS

luisacherrera:


starter-code/index.html
```diff
@@ -2,41 +2,94 @@
...
+      Spotify is now free on mobile, tablet and computer.
+      <br/>
+      Listen to the right music, wherever you are.
+    </p>
+  </div>
+  <div class='definition'>
+    <div class='what'>
+      <h2>What’s on Spotify?</h2>
+    </div>
+    <div class='million'>
```
tawebbcn: You can name all three div tags with the same class so that you can reuse css code instead of styling three times the same.

starter-code/css/main.css
```diff
@@ -6,3 +6,211 @@ Green: #00B172
...
+
+html {
+    box-sizing: border-box;
+  }
+  *, *:before, *:after {
+    box-sizing: inherit;
+  }
+  
+  
+
```
tawebbcn: you have to reset the body too. If you inspect the body tag in your browser, you'll see that it has an ugly white thin margin. So you have to set body margin an padding to 0.

PR #298, opened 2018-01-11
BCN Martin

martinX308:


starter-code/index.html
```diff
@@ -3,40 +3,80 @@
 <head>
   <meta charset="UTF-8">
   <title>Spotify</title>
+  <link rel="stylesheet" href="./css/main.css">
 </head>
 <body>
-Premium Discover Help Download
+  <div class="spotify-menu">
```
tawebbcn: Be careful on naming. All your html is wrapped by a container called spotify-menu. Better name it 'main-container/wrapper'

starter-code/css/main.css
```diff
@@ -6,3 +6,151 @@ Green: #00B172
...
+      position:absolute;
+      top:0;
+      width:60%;
+  }
+  .header-left{
+      float: left;
+      width:30%;
+      padding:1em;
+  }
+  .header-right{
```
tawebbcn: try not to name classes with their position.

starter-code/css/main.css
```diff
@@ -6,3 +6,151 @@ Green: #00B172
...
+  }
+
+  .buttom-element div{
+      float: left;
+  }
+  .buttom-text-float{
+    width:50%;
+  }
+  .app{
+      height:22em;
```
tawebbcn: It's ok to use relative units, but better use em and rem for styles related to fonts. To style divs and images better use vh, vw, %... Also, it is better to set images with absolute sizes (like pixels).

Check this little article about relative units and their meaning:

PR #300, opened 2018-01-11
BCN - Dani 

DTDK:
not much to commit...

### DIFF
--- a/starter-code/css/main.css
+++ b/starter-code/css/main.css
@@ -6,3 +6,58 @@ Green: #00B172
 White: #FFF
 
 */
+
+html{
+ 
+}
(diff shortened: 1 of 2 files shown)


starter-code/css/main.css
```diff
@@ -6,3 +6,58 @@ Green: #00B172
...
+    padding: 0 5px;
+}
+
+.center-container {
+    clear: both;
+    color: white;
+    background: url(../images/landing.jpg) no-repeat;
+    overflow: hidden;
+    border: 1px solid red;
+
```
tawebbcn: You have to set the div height. Also try 
```
background-size: cover
```

PR #301, opened 2018-01-11
BCN - Filipe Rainho

ncoder92:


starter-code/css/main.css
```diff
@@ -5,4 +5,199 @@ Text: 1A1A1A
...
+
+.spotify-logo {
+    width: 200px;
+    margin: 10px 20px 10px 30px;
+    animation: 2s linear 2s linear;
+    animation-direction: alternate;
+    position: relative;
+}
+
+.spotify-logo:hover {
```
tawebbcn: Hahahah super nice bonus!

starter-code/index.html
```diff
@@ -3,40 +3,58 @@
...
+      <div id='landing-text'>
+        <h1>Music for everyone.</h1>
+        <p>Spotify is now free on mobile, tablet and computer.</p>
+        <p>Listen to the right music, wherever you are.</p>
+      </div>
+    </div>
+    <div id='features'>
+      <p class='main-features-title'><span>What’s on Spotify?</span></p>
+      <div class='main-features'>
+        <div id='feature1'>
```
tawebbcn: Set the same class for the three div tags and you will be able to reuse the code instead of writing the same styling for all of them

PR #302, opened 2018-01-11
BCN - David

davidalexteixeira:
still to be completed - will complete over the weekend - had issues with getting the three icons to place properly.

### DIFF
--- a/starter-code/css/main.css
+++ b/starter-code/css/main.css
@@ -57,24 +57,77 @@ li {
 .image-wrapper {
     
     clear: both;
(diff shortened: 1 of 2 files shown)


starter-code/css/main.css
```diff
@@ -6,3 +6,219 @@ Green: #00B172
...
+}
+  
+  /* --------- layout ---------- */
+  
+  .main-wrapper{
+      margin-top: 30px;
+      margin-left: 30px
+  }
+  
+
```
tawebbcn: Try to fill better the layout section.

Layout is like the skeleton.

PR #303, opened 2018-01-11
BCN Pedro Neto

pedrodiogoneto:


starter-code/css/main.css
```diff
@@ -6,3 +6,122 @@ Green: #00B172
...
+    font-size: 25px;
+}
+
+.nav-options {
+    overflow: hidden;
+    background-color: white;
+}
+
+.background-picture {
+    background-image: url(/Users/pedroneto/Desktop/ironhack/week1/lab-css-spotify-clone/starter-code/images/landing.jpg);
```
tawebbcn: RELATIVE PATHS!!! It should look like this:
```
```

starter-code/css/main.css
```diff
@@ -6,3 +6,122 @@ Green: #00B172
...
+    color: green;
+}
+
+.images-advertising {
+    height: 50%;
+    width: 50%;
+}
+
+.advertising-container {
+    width: 400px;
```
tawebbcn: Try to use more relative units in sizing. Instead of using pixels, try to use %, vh, vw, em...

PR #309, opened 2018-01-11
BCN Jan Web Dev FT - Matt ///

mattmegarry:


starter-code/css/main.css
```diff
@@ -1,3 +1,46 @@
...
+    .nav-buttons {
+        float: right;
+        }
+
+    .nav-buttons ul li{
+        display: inline;
+        /* margins needed */
+    }
+
+/* Where to put this?  */ ul {list-style: none;}
```
tawebbcn: Try not to style element tags like div or ul. Give them a class or ID instead.

You can put it in components :)

starter-code/css/main.css
```diff
@@ -1,3 +1,46 @@
+/* reset */
+
+/* typography  ...including h1 margins and padding etc */
+
+/* layout */
+    .main {
+        right: 0;
+        left: 0;
```
tawebbcn: These are useless.

starter-code/css/main.css
```diff
@@ -1,3 +1,46 @@
+/* reset */
+
```
tawebbcn: You have to copy here a css reset

starter-code/css/main.css
```diff
@@ -1,3 +1,46 @@
...
+/* layout */
+    .main {
+        right: 0;
+        left: 0;
+        width: 100%; 
+    }
+
+    .navbar {
+        position: fixed;
+        right: 0;
```
tawebbcn: Same here! You can remove the right/left: 0. What you have to do is to reset the css (body has paddings by deffault and that's what you got a little white border there)

starter-code/css/main.css
```diff
@@ -6,3 +49,8 @@ Green: #00B172
 White: #FFF
 
 */
+
+/* --------- TEMP ----------*/
+div, header, footer, nav, article {
+	border: 1px solid black;
```
tawebbcn: better set them red

starter-code/css/main.css
```diff
@@ -1,3 +1,46 @@
...
+    }
+
+/* Where to put this?  */ ul {list-style: none;}
+
+/* pages */
+
+
+
+body {
+    color: red;
```
tawebbcn: why font color red?

PR #313, opened 2018-01-11
BCN - Sheng

theRealSheng:


starter-code/css/main.css
```diff
@@ -6,3 +6,184 @@ Green: #00B172
...
+.top-picture{
+    background-image: url(../images/landing.jpg);
+    background-repeat: no-repeat;
+    width: 100%;
+    background-size:1250px 650px;
+    margin-top: 100px;
+    padding-bottom: 190px;
+}
+
+h1{
```
tawebbcn: All this styles go into Typography section

starter-code/css/main.css
```diff
@@ -6,3 +6,184 @@ Green: #00B172
...
+}
+
+.bottom-middle{
+    width:300px;
+    margin-left: 200px;
+}
+.bottom-circle{
+    width:40%;
+    margin-left: 300px;
+    margin-top: -310px;
```
tawebbcn: Maybe just set the image as a background and give it size. Instead of this tricky thing

PR #316, opened 2018-01-11
BCN / Oscar Jorge ---  working on this

oscarjorgejorge:


starter-code/css/main.css
```diff
@@ -6,3 +6,100 @@ Green: #00B172
...
+  margin-left: 10px;
+}
+
+li {
+  list-style: none;
+  padding-right: 20px;
+
+}
+
+nav {
```
tawebbcn: Don't use element tags to style. Better add them a class or ID

PR #325, opened 2018-01-24
[MIA] Mosi Rivera

mosileko:


jessicaalin:
Solid work! I would just recommend using more scaleable units and some comments in your CSS of the sections. Sometimes a class like "wannabe" isn't too descriptive :P

PR #326, opened 2018-01-24
[MIA] [Stephen-Jackson]

stephjackson:
did a second commit to fix text color/logo animation, check that one.

### DIFF
--- a/starter-code/css/main.css
+++ b/starter-code/css/main.css
@@ -6,3 +6,152 @@ Green: #00B172
 White: #FFF
 
 */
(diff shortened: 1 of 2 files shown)


jessicaalin:
Great job! A couple of notes:
1. Careful with the absolute for the last section. Even though this doesn't have to be responsive, using that positioning made the what-container too long. Also, missing some margin on this section. 
2. Better to link your Google font in your HTML than in CSS using import as that CSS rule will defer loading.

PR #327, opened 2018-01-24
[MIA] [Nicolas - Levo]

nlevo:


jessicaalin:
Great job! I would recommend you to add `body {margin: 0 auto;}` to get rid of that extra white border around the page (that shows on the landing image).

PR #328, opened 2018-01-24
[MIA][Scott-Rowell]

rowellscott:


jessicaalin:
Solid work! I would just recommend for you to comment in your CSS a bit what the sections are and for you to use scaleable units (you start using them but can be more consistent).

PR #329, opened 2018-01-24
[MIA] [Jeremy Fox] - spotify clone 

jfoxxxx:


jessicaalin:
All right, good job on getting the overall layout. A few notes:
1. Your middle section is almost there. Had you floated each column, they layout for the middle would have been good. (The right-col you set `margin: right`, should have been `float: right`.)
2. After setting this, you have to add the clear-fix or the rest of your elements will float up to the available space. 
3. Another quick tip, try setting `body { margin: 0 auto } as it'll get rid of that extra white space that borders the page.

PR #331, opened 2018-01-25
[MIA] [Jonathan - Sanchez]

jonathanlsanch:


jessicaalin:
Great job! Really good attention to detail.
Okay, somethings:
1. With some more playing around of the Kanye image and the Spotify logo, you would have got it.
2. Notice that around the landing image there is a little white border? If you apply `body { margin: 0 auto} it'll take it away.

PR #332, opened 2018-01-25
[MIA] Rakeem-Thomas Spofity Clone

rthomas29:
Implementing the layout with Grid definitely made the process more intuitive.


jessicaalin:
Great job! You get a star! Jk. We don't do those... However,  I would like to tell you to be super duper careful with that universal selector. This will be a PITA in bigger projects. (Think of components that live in components that live components...) Instead of adding those properties to the universal, just add it to the body. In this case it will render the same results.

PR #336, opened 2018-01-25
[MIA] [GYASIK - NUNEZ]

gyasikn:
need to finish center section. top and bottom are done for now.

### DIFF
--- a/starter-code/css/main.css
+++ b/starter-code/css/main.css
@@ -6,3 +6,133 @@ Green: #00B172
 White: #FFF
 
 */
+body {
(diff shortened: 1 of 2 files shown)


jessicaalin:
Nice! Solid start. Just a few notes:
1. Your Kanye image scrolls over your nav-bar. You just need to add a z-index to your nav-bar to fix that. 
2. Some padding on the Spotify logo in the nav and smaller in size too.
3. Your HTML structure in terms of indentation is a little off. I would recommend opening the dev tools and seeing the DOM. Indentation helps to notice the relationship between elements (parent/child) or also for readability (going on a new line and indenting content). Siblings should have the same indentation.

PR #337, opened 2018-01-25
[MIA] {Daniel-Autrey]

Dautrey94:
Had trouble with the lower part.
could not position them correctly


jessicaalin:
Really good job! With a little more playing on the padding and positioning you would have got it. Good job on the sticky nav-bar. Just a few notes:
1. You might want to add something line `body { margin: 0; }` so that way you can clear out the white border around the page.
2. Be careful of using inline styles. (Like how you set the width and height in the HTML.) It is better to style with CSS as inline styles override an external styles. To override inline styles, you would need to use !important in CSS.
3. Look at how your float-3 image is stretched incorrectly. This is because you specified both the height and the width; if you just specify one, then the other will be auto and preserve the image's dimensions.

PR #339, opened 2018-01-25
[MIA] LEXY-VANDERFORD

LexyV:
//wasn't able to complete, still working on the tedious details but wanted to submit what I had so far.

### DIFF
--- a/starter-code/css/main.css
+++ b/starter-code/css/main.css
@@ -6,3 +6,105 @@ Green: #00B172
 White: #FFF
 
 */
+
+*{
+    font-family: open sans;
+}
+.nav-bar{
+    position: fixed;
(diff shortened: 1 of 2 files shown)


jessicaalin:
Good job! Just a few things:
1. Nav-bar has a few bugs. So your elements list items are a little off. You could have easily used `display: inline;` instead of position or left to get them to line up correctly. Also, the logo is not included in your nav-bar structure so that it is not fixed. Lastly, even though the nav-bar is set to a fixed position, you need to give it a z-index so that it will remain on top of the next element (being the header in this case).
2. I would recommend you to set the landing image as a background image on a div so it is easier to manipulate since it is has text on top of it.
Definitely on the right track!

PR #340, opened 2018-01-25
[Mia] [Michael - Elso]

ElsoMichael:
Issue: 
-Had issue randomly with fixed position nav bar, it went behind everything so i had to push it up with a z-index, but my issue was why it suddenly with out no reason pushed behind everything.


jessicaalin:
Off to a good start. I was playing around to see the error you were getting, so I took off the z-index. It's still actually there if you scroll down - it's just the image that covers it. And that's because when we set an element to a fixed position, it takes it out of flow. The next element you have is set to a relative position, so it behaves as if it were the first along and that's why you need the z-index.

A few other notes: 
1. For the fixed nav-bar, you should set a top to it because you can see elements are just scrolling behind it.
2. Right now you have the ability to scroll horizontally (bad UX), so you might want to try to set width to 100% somewhere to set it to the width of the page.
3. Some more work on the positioning.

PR #341, opened 2018-01-25
[MIA][Ana-Raballo]

AnaRaballo:
Need to review positioning and box model on css.

### DIFF
--- a/starter-code/css/main.css
+++ b/starter-code/css/main.css
@@ -6,3 +6,70 @@ Green: #00B172
 White: #FFF
 
 */
+
+body {
+    font-family: "Open Sans";
+}
+
+h1 {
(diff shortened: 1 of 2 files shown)


jessicaalin:
Off to a good start. I would recommend you to organize when it comes to creating a page. You can start by getting the entire structure of the page done and then start working on the styling, or finalize one part of the structure (like the nav-bar) and then do the styling for that. 

Another thing is that when you're working with images that have overlay, it's better to set these images as a background image for a container (that will hold the text). That way you have access to properties such as background-size while setting that container to 100% of the page and still preserving the photo. Right now your page can scroll horizontally (bad UX) because of the set dimensions of the photo.

PR #352, opened 2018-01-25
[MIA] Pablo - Gomez

ZTylerDurden:
Issues:

Did not finish but was able to create the first two sections somewhat well.  If I had more time I'd continue, possibly this weekend.

### DIFF
--- a/starter-code/css/main.css
+++ b/starter-code/css/main.css
@@ -6,3 +6,73 @@ Green: #00B172
 White: #FFF
 
 */
+
+.nav-bar {
+    background: white;
+    height: 97px;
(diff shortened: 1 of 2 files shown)


jessicaalin:
All right, off to a good start. Some things to note:
1. Should have set a top value for the fixed nav-bar since right now your image can be seen scrolling passed it.
2. The image is too wide and is allowing you to scroll sideways on the page (bad UX). What you could have done with the image is set it as the background to a div, which would give you access to properties such as background-size which you can set to cover or contain. This let's you set the div to 100% of the page and would preserve the photo still.

PR #353, opened 2018-01-25
[MIA][ALEX VIRDEE]

alexvirdee:


jessicaalin:
Needs work with the padding and everything is on a smaller scale. So keep in mind to use the space. Also, be super careful with this: 
`* {
    margin: 0;
    padding: 0;
}`
Universal selector blocks inheritance and gives everything single element that style. So, properties that you imagine an element will inherit from their parent will not. The selector is cool for fonts, but not so much margins.

PR #359, opened 2018-03-21
BCN - Jordi & George

jordicalzada:
@GeorgeCoupland


starter-code/index.html
```diff
@@ -2,41 +2,86 @@
...
+  <nav id="nav-bar" class="container">
+    
+    <img src="images/spotify-logo.png">
+    
+    <ul>
+      <a href="#" ><il>Premium</il></a>
+      <a href="#"><il>Discover</il></a>
+      <a href="#"><il>Help</il> </a>
+      <a href="#"><il>Download</il></a>  
+  </ul>
```
tawebbcn: Under a ```ul``` tag you should use ```li``` tags as siblings and not wrap them around other elements. You used ```il``` which is an invalid tag so be careful.

starter-code/index.html
```diff
@@ -2,41 +2,86 @@
...
+  </nav>
+
+  <section class="landing">
+    <div class="text">
+      <h1><b>Music for everyone.</b></h1>
+      <h2><c>Spotify is now free on mobile, tablet and computer.<br>
+        Listen to the right music, wherever you are.</c></h2>
+    </div>
```
tawebbcn: There's a c tag in here :eyes:

starter-code/index.html
```diff
@@ -2,41 +2,86 @@
...
+    Just search and hit play.</p>
+<h3>Browse</h3>
+<p>Check out the latest charts,
+    brand new releases and great
+    playlists for right now.</p>
+<h3>Discover</h3>
+<p>Enjoy new music every Monday
+    with your own personal playlist.
+    Or sit back and enjoy Radio.</p>
+</div>
```
tawebbcn: Don't forget to ident your code.

PR #362, opened 2018-03-21
BCN Fabian-N Ioana-D

IoanaOrca:
The best we could do!


starter-code/css/main.css
```diff
@@ -6,3 +6,200 @@ Green: #00B172
...
+    box-sizing: border-box;
+  }
+  *, *:before, *:after {
+    box-sizing: inherit;
+  }
+  
+  body {
+    margin: 0;
+    padding: 0;
+  }
```
tawebbcn: Reset :+1:

starter-code/css/main.css
```diff
@@ -6,3 +6,200 @@ Green: #00B172
...
+
+ .second-part{
+     margin:5%;
+     padding:5px 10px;
+     overflow:hidden;
+ }
+.second-part h2{
+    color:#1a1a1a;
+    text-align:center;
+    text-decoration: underline;
```
tawebbcn: Nice choice! Border bottom could also be used :+1:

PR #367, opened 2018-03-21
BCN - Lennie - Pere

lenniez:
We are almost there! Turning it in now so we can move on to the JS exercise. CSS IS AWESOME. @Interna1ta

### DIFF
--- a/starter-code/css/main.css
+++ b/starter-code/css/main.css
@@ -1,3 +1,153 @@
+/* ---- reset ---- */
+
+html {
+  box-sizing: border-box;
+}
+*, *:before, *:after {
+  box-sizing: inherit;
+}
+
+body {
+  margin: 0;
+  padding: 0;
+}
+
+/* ----layout ----*/
+
+#site-header {
+  height: 100vh;
+  background-image: url("../images/landing.jpg");
+  background-size: cover;
+  background-position: center;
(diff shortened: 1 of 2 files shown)


starter-code/css/main.css
```diff
@@ -1,3 +1,153 @@
...
+  line-height: 100px;
+}
+
+.icon-desc p {
+  font-size: 30px;
+  color: #6E6E6E;
+  font-weight: 100;
+}
+
+#kanye {
```
tawebbcn: better do it with a class, because it shuld be reusable for other authors

starter-code/css/main.css
```diff
@@ -1,3 +1,153 @@
...
+  width: 200px;
+}
+
+.icon-desc {
+  display: inline-block;
+  margin: 30px;
+  width: 20%;
+}
+
+#kanye {
```
tawebbcn: missing background white logo image

starter-code/css/main.css
```diff
@@ -1,3 +1,153 @@
...
+#landing-text {
+  padding: 20vh;
+}
+
+.whats-on-img { 
+  width: 200px;
+}
+
+.icon-desc {
+  display: inline-block;
```
tawebbcn: :+1:

PR #371, opened 2018-03-21
[MAD DE] Pablo García

HeyPablete:
He terminado todo
Mi mayor dificultad está en cuando usar el display: inline-block o el inline-flex. En el footer había colocado un display: inline-flex al div padre de la imagen de la app y la imagen se deformaba, lo cambié por un display: inline-block y se solucionó.


HeyPablete:
Animación  añadida al logo

ta-web-mad:
Aquí te dejo la diferencia entre ambos un saludo!

display: inline-flex does not make flex items display inline. It makes the flex container display inline. That is the only difference between display: inline-flex and display: flex. A similar comparison can be made between display: inline-block and display: block, and pretty much any other display type that has an inline counterpart.1

There is absolutely no difference in the effect on flex items; flex layout is identical whether the flex container is block-level or inline-level. In particular, the flex items themselves always behave like block-level boxes (although they do have some properties of inline-blocks). You cannot display flex items inline; otherwise you don't actually have a flex layout.

It is not clear what exactly you mean by "vertically align" or why exactly you want to display the contents inline, but I suspect that flexbox is not the right tool for whatever you are trying to accomplish. Chances are what you're looking for is just plain old inline layout (display: inline and/or display: inline-block), for which flexbox is not a replacement; flexbox is not the universal layout solution that everyone claims it is (I'm stating this because the misconception is probably why you're considering flexbox in the first place).

PR #394, opened 2018-03-22
PAR Bertrand

wavilo:
I have to go


ta-web-paris:
Good job, you manage to play with block and to align them right ! 
The only small thing I can say is about the navbar elements, a small padding would have been great :) 

Otherwise 👍

PR #395, opened 2018-03-22
PAR Mostafa Gad

M-Gadd:


starter-code/css/main.css
```diff
@@ -6,3 +6,132 @@ Green: #00B172
 White: #FFF
 
 */
+.body{
```
ta-web-paris: to add font (and make your website look better ;) ) you can use this piece of code, basically it's calling a google font and it allow you to use it anywhere you want on your website : 

```css

@import url("https://fonts.googleapis.com/css?family=Raleway");
@import url("https://fonts.googleapis.com/css?family=Montserrat:400,700");

* {
  margin: 0;
  padding: 0;
  font-family: MOntserrat;
}

```

starter-code/index.html
```diff
@@ -3,28 +3,58 @@
...
+<p>Spotify is now free on mobile, tablet and computer. <br>
+  Listen to the right music, wherever you are.</p> 
+</div>
 
-What’s on Spotify?
+<div class="secPage">
 
-Millions of Songs
-There are millions of
+<header>
```
ta-web-paris: You should use a div instead, and call it the way you want, but usually the header is on the top of your webpage

ta-web-paris:
I know it's not the more intersting part, but it's really important for you to be ok with HTML & CSS as you are going to use later to build awesome and dynamic website ! ;) 

Don't give up ! and if you need help on some part just tell me

PR #397, opened 2018-03-22
PAR Louis 

LouisMasurel:
Details details details...

### DIFF
--- a/starter-code/css/main.css
+++ b/starter-code/css/main.css
@@ -1,8 +1,146 @@
 /*
 Colors:
 
-Text: 1A1A1A
+Text: #1A1A1A
 Green: #00B172
 White: #FFF
 
 */
+
+body {
+ text-decoration: none;
+
+}
+.navbar {
+  display: inline;
+  text-align: center;
+  font-size: 20px;
+}
+
+.spotify {
+  height: auto;
+  width: 300px;
+  margin-right: 600px;
+}
+
+.concert {
+  background-image : url("../images/landing.jpg");
+  background-repeat: no-repeat;
+  height: 300vh;
+  width: 100vw;
+  background-size: cover;
+  text-align: center;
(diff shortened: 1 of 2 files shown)


starter-code/css/main.css
```diff
@@ -1,8 +1,146 @@
...
 Green: #00B172
 White: #FFF
 
 */
+
+body {
+ text-decoration: none;
+
+}
+.navbar {
```
ta-web-paris: Try to use this code (but of course adapt it to your class and structure) it should help you to have a nice navbar

```css

#header {
  position: fixed;
  height: 90px;
  background-color: white;
  width: 100%;
  display: flex;
  box-shadow: 0px 10px 5px -8px rgba(0, 0, 0, 0.75);
}

#header img {
  height: 70%;
  margin: 10px 0 0 30px;
}

#header ul {
  display: flex;
  align-items: baseline;
  margin: 2% 0 0 50%;
}

#header li {
  margin: 0 20px;
  list-style-type: none;
}

```

starter-code/index.html
```diff
@@ -1,42 +1,88 @@
...
+    <section class="container section left">
+          <h6>Search</h6>
+          <p>
+            <em>Know what you want to listen to?</em>
+            <em>Just search and hit play.</em>
+          </p>
+
+          <h6>Browse</h6>
+          <p>
+            <em>Check out the latest charts,</em>
```
ta-web-paris: You shouldn't use em like this, you should use p tag instead.
This em tag is supposed to highlight piece of text in a big text per exemple.

try stay simple and use only what you need.

ta-web-paris:
Next time before rushing into the code, take 10minutes to draw your man block and all your elements so it will be easier for you to build your structure and to style it :)

don't give up !

PR #398, opened 2018-03-22
PAR Lucas

lucasm-git:
It looks OK on my computer, but the responsiveness isn't quite good yet so I'll try to think of it before the next CSS exercise.

### DIFF
--- a/starter-code/css/main.css
+++ b/starter-code/css/main.css
@@ -6,3 +6,220 @@ Green: #00B172
 White: #FFF
 
 */
+
+
+/*****************************************************************
+        RESETS
+*****************************************************************/
+
+
+html, body, div, span, applet, object, iframe,
+h1, h2, h3, h4, h5, h6, p, blockquote, pre,
+a, abbr, acronym, address, big, cite, code,
+del, dfn, em, img, ins, kbd, q, s, samp,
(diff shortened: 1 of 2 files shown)


starter-code/css/main.css
```diff
@@ -6,3 +6,220 @@ Green: #00B172
...
 
 */
+
+
+/*****************************************************************
+        RESETS
+*****************************************************************/
+
+
+html, body, div, span, applet, object, iframe,
```
ta-web-paris: Nice ! 

It's good to reset the default style of the browser before building your website :)

starter-code/css/main.css
```diff
@@ -6,3 +6,220 @@ Green: #00B172
...
+
+/*****************************************************************
+        HEADER
+*****************************************************************/
+
+header {
+    height: 908px;
+    width: 100%;
+    background-image: url("../images/landing.jpg");
+    background-size: contain;
```
ta-web-paris: On my computer it looks like that your "contain" is not the best choice, when I change it to cover (what I would recommand you) it's better (as the image is now full screen).

But it's the only things I have to say, otherwise it's a really good job !

ta-web-paris:
Good job Lucas, keep working hard ! 👍

PR #399, opened 2018-03-22
PAR Renaud

Ayss83:


starter-code/css/main.css
```diff
@@ -6,3 +6,137 @@ Green: #00B172
...
 
 */
+* {
+  border: solid transparent 1px;
+}
+
+body {
+  font-family: 'Montserrat', sans-serif;
+  margin: 0 auto;
+  width: 1400px;
```
ta-web-paris: When it comes to your body, don't use absolute value like px, because as soon as you will open the website on an other computer it may not respond as expected, instead try with a 100% :)

ta-web-paris:
Good job Renaud, keep working hard 🥇

PR #400, opened 2018-03-22
PAR ELISE

elisegriset92:
merci Vivian !!
Pas compris pourquoi mon trio ne s'aligne pas au niveau des titres mais ok pour le reste !

### DIFF
--- a/starter-code/index.html
+++ b/starter-code/index.html
@@ -3,40 +3,99 @@
 <head>
   <meta charset="UTF-8">
   <title>Spotify</title>
+  <link rel="stylesheet" href="../starter-code/css/main.css">
+  <link href="https://fonts.googleapis.com/css?family=Montserrat" rel="stylesheet">
+
 </head>
 <body>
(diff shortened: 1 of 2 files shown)


starter-code/index.html
```diff
@@ -3,40 +3,99 @@
...
-brand new releases and great
-playlists for right now.
-Discover
-Enjoy new music every Monday
-with your own personal playlist.
-Or sit back and enjoy Radio.
+<header class="headerBackground">
+
+<div class="img">
+  <img src="../starter-code/images/spotify-logo.png" width="400px" alt="spotify-logo"></div>
```
ta-web-paris: May be you can try with a smaller width for your logo.
Also try not to use inline css like this, it's better to keep all you styles on a dedicated file :)
And it will be easier for you to change it !

This is the only point I have to mention, otherwise good job :)

ta-web-paris:
Good job Elise ! 👍

PR #401, opened 2018-03-22
PAR Jb 

Jbmouterde:


starter-code/index.html
```diff
@@ -3,40 +3,82 @@
...
+
+  <h3>Spotify is now free on mobile, tablet and computer. <br>
+      Listen to the right music, wherever you are.</h3>
+ </div>
+
+
+ <section>
+  <h3> <strong> What’s on Spotify? </strong>  </h3>
+  <div class="songs">
+      <img src="./images/music-icon.png" height="140" width="140">
```
ta-web-paris: Actually it's not a good practice to put inline style like this, and as you are applying the same style on your three elements, you should use class and put it in your css file.

Also small trick when it comes to resize image, use a width like you did and use auto for you height.

```css

#green-container-img img {
  width: 60%;
  height: auto;
  float: right;
}

```

This is the code I use for this exercise, like this your image will keep the same proportions.

ta-web-paris:
Good job, small details but you are on the good way :)

Keep up the good work 👍

PR #402, opened 2018-03-22
PAR Jeremie

JeremieHercberg:


ta-web-paris:
Don't give up ! It may not be the most interesting part, but it's super important as it will help you build awesome website later during the bootcamp !

If you need any advice, or help we are here to help you :)

PR #405, opened 2018-03-22
PAR Leo P.

LPsola:
Couldn"t manage to align colomns properly in the What's on Spotify section


ta-web-paris:
Actually it's a pretty good job ! You will see flexbox next week (or later today) it will help you a lot with block alignment :)

PR #409, opened 2018-03-23
PAR Rodrigo

LazerLight:
I only completed half. I thought we were going to get more time today to complete it.


ta-web-paris:
No worries :)
It's almost perfect.
I think with more time you could have been finish the whole exercise.

PR #411, opened 2018-03-23
PAR Sami

Sami-Az:
Still didnt finish it but any feedbacks would be appreciated


ta-web-paris:
The markup is clean.
On the bottom part why did you split all the boxes by section and specific class?
Just on class is good.
After you can add:
```cs
width: 36vw; 
```
_for example..._

This is my work:
https://github.com/VivianSolide/lab-css-spotify-clone

I did the landing page easily thanks to flex concept.

PR #414, opened 2018-04-04
MIA-FTWD-DiegoLeal-SpotifyClone

DiegoLealB:


starter-code/css/main.css
```diff
@@ -6,3 +6,88 @@ Green: #00B172
 White: #FFF
 
 */
+body{
```
ianizaguirre: When we create a website WE MUST DEFINE some sort of container width for the whole page. We can choose to add the following CSS rule to the `<body>` tag or choose to add this rule to a `<div class="wrapper"> element that we add after our opening body tag and close before our closing body tag. So we have to center our page, lets say we choose to add the rule to the body tag, then we should add a width and margin...
```css
body {
   width: 100%;
   margin: 0 auto;
}
```
.... Try adding that to this project, and see if you notice the difference? You will see a BIG BIG difference ... your page will be centered and your content will fill the browser exactly how you want it.  😁

ianizaguirre:
Try to apply this feedback on your next project. 

Kind Regards,
Ian Izaguirre | Ironhack TA

PR #415, opened 2018-04-04
MIA - FTWD - Richard Batista

blue-stalfos:


starter-code/index.html
```diff
@@ -1,42 +1,102 @@
...
+		<div class="main-before-insert">
+
+			<div class="main-whats-on-spotify">
+				What’s on Spotify?
+			</div>
+			
+			<div class="feature-containers">
+
+				<div class="main-feature-1 float-l">
+					<img src="images/music-icon.png" width="10.5%">
```
sandrabosk: You need this text to go under the image and span makes it in the same line so maybe you're better off putting it in the p tag or div tag.

starter-code/index.html
```diff
@@ -1,42 +1,102 @@
...
+					computer
+				</div>
+
+			</div>
+		</div>
+
+		<div class="clearfix"></div>
+
+		<div class="main-body-insert">
+			<div class="left-text">
```
sandrabosk: Maybe it would be more practical if you divided this section into two divs (on that holds everything that's on the left side and the other one that holds everything that's on the right side) and after that just use the classes to position them.

sandrabosk:
Good work so far. Thank you for submitting.

PR #421, opened 2018-04-05
MIA-JEYSON-SPOTIFY_CLONE

jmxxii:


starter-code/css/main.css
```diff
@@ -6,3 +6,176 @@ Green: #00B172
...
+  font-size: 25px;
+  text-align: left;
+  color: white;
+  font-weight: 100;
+  padding-left: 3.5em;
+  line-height: 1.5;
+}
+
+#kanye {
+  float: left;
```
ianizaguirre: You used the `float:left` property in a bunch of CSS classes ... When you start writing more CSS and you notice that you are repeating something over and over again, then its best to make it into its own class... So for example, you could have created a class called something like:
```css
.column-left {
  float: left;
}
```
and  then deleted `float:left` from all your classes in your HTML and then in your html you could have added `column-left` to the classes that would need a float left... for example:
```html
<div id="stream" class="column-left">
```
Or you could have made your life easier and decided to not use any #ids and just use classes instead, so then you could just have:
```html
<div class="column-left stream">
```
For now, Stick with CSS classes. They are reusable and their specificity can be easily reused.

Styling IDs goes against DRY principles and is something that should be avoided, if it makes sense. 


There are situations where you need to style a single object and you know it will never be shared with anything else in those rare case we will use ID, as long as you're aware of the maintainability risk and it's acceptable do what you need to do to get the job done. 

But in a situation like this , especially when learning , I think we should just not use it , and you used them WAY WAY to much in your HTML.

starter-code/index.html
```diff
@@ -1,42 +1,91 @@
...
+                Know what you want to listen to?<br/>
+                Just search and hit play.
+              </p>
+            <h4 class="kanyeh4">Browse</h4>
+              <p>
+                Check out the latest charts,<br/>
+                brand new releases and great<br/>
+                playlists for right now.
+              </p>
+            <h4 class="kanyeh4">Discover</h4>
```
ianizaguirre: I would not write something like `"kanyeh4"` ... adding the `h4` part to the class, is what I don't like. Choose a better name, something like: `"kanye-title"`

ianizaguirre:
Try to apply this feedback on your next project.

Kind Regards,
Ian Izaguirre | Ironhack TA

PR #427, opened 2018-04-05
mia-ftwd-jens heilemann-spotify clone

jkai7:
no pics yet

### DIFF
--- a/starter-code/css/main.css
+++ b/starter-code/css/main.css
@@ -6,3 +6,110 @@ Green: #00B172
 White: #FFF
 
 */
+body {
+    font-family: Helvetica Neue, Sans-Serif;
+}
+
+  /* nav bar */
+
+.nav-bar {
+    position: fixed;
+    top: 0;
+    width: 99%;
+    height: 83px;
+    background-color: white;
+}
+.nav-bar .links ul {
+       float: right;
+       list-style: none;
+       display: flex;
+        margin-top: 35px;
+        margin-right: -1%;
+        color: #1A1A1A;
+}
+.nav-bar .links ul li{
+    padding-right: 50px;
+}
+.logo img {
+    float: left;
+    height: 35px;
+    width: 110px;
+    margin-top: 30px;
+    margin-left: 3%;
+}
+.clear {
+      clear: both;
+}
+
+/* everyone */
+
+.everyone {
+    margin-top: 83px;
+    background-image: url(../images/landing.jpg);
+    background-size: 100% 100%;
+    height: 600px;
+    color: white;
+}
+.everyone h1 {
+    text-align: center;
+    font-size: 4.9em;
+    padding-top: 130px;
+}
+.everyone p {
+    width: 52%;
+    margin-left: 24%;
+    text-align: center;
+    font-size: 2em;
+}
+
+/* whats-on */
+
+.whats-on {
+    margin: 100px 20px 230px 0px;
+}
+
+.whats-on h2 {
+    text-align: center;
+    border-bottom: 3px solid #00B172;
+    width: 16%;
+    line-height: 1.5;
+    margin-left: 590px;
+    color: #1A1A1A;
+}
+.flex-container {
+    display: flex;
+}
+.whats-on div {
+    float: left;
+    width: 100%;
+    text-align: center;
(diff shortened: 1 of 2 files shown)


starter-code/index.html
```diff
@@ -3,40 +3,90 @@
...
 
-Music for everyone.
-Spotify is now free on mobile, tablet and computer.
-Listen to the right music, wherever you are.
+<!-- nav bar -->
 
-What’s on Spotify?
+<div class="nav-bar">
+  <div class="logo"><img src="./images/spotify-logo.png" alt="logo"></div>
+<div class="links">
```
ianizaguirre: Please start getting into the habit of properly spacing/tabbing your divs, because all of your divs are on the same left side as the body tag, so now it is hard to follow where starting and closing tags start and end.

starter-code/index.html
```diff
@@ -3,40 +3,90 @@
...
+<h2>What’s on Spotify?</h2>
+
+<div class="flex-container">
+<div>
+<h3>Millions of Songs</h3>
+<p>There are millions of
+songs on Spotify</p>
+</div>
+
+<div>
```
ianizaguirre: There is no need to make a div class without a name. Dont make empty `<div>` without giving a class name.

starter-code/index.html
```diff
@@ -3,40 +3,90 @@
...
-It’s as yeezy as Kanye West.
-Search
-Know what you want to listen to?
-Just search and hit play.
-Browse
-Check out the latest charts,
+computer</p>
+</div>
+
+</div><!--flex container-->
```
ianizaguirre: Great Job on leaving HTML comments on your code 👍

starter-code/index.html
```diff
@@ -3,40 +3,90 @@
...
+</div>
 
-HD Music
-Listen to music as if you
-were listening live
+<!-- whats on spodify -->
 
-Stream Everywhere
-Stream music on your
+<div class="whats-on">
```
ianizaguirre: I do not like your choice for this div name... instead of something so general like `"whats-on" , a better choice would be something like "options-wrapper" , or "options-container", or "middle-links" or "middle-row" ..... the point is , do not pick to name a div class by choosing some words that are part of that section, like you chose "whats-on" because the heading of that section is called "what's on Spotify" .  The reason to avoid doing this, is because usually we are going to reuse div names throughout our code, so if we were to stick with something so specific to that section, like sticking with calling it "whats-on", then it would be messy to reuse this name in another part of your code , since there probably only be one section using that heading "whats-on" , if we were to pick a better name then it would fit to reuse it in other parts of the code.

starter-code/index.html
```diff
@@ -3,40 +3,90 @@
...
+  </ul>
+</div>
+</div>
+<div class="clear"></div>
+<!-- music for everyone -->
 
-Millions of Songs
-There are millions of
-songs on Spotify
+<div class="everyone">
```
ianizaguirre: I do not like your choice for this div name "everyone"... ( see my comment below about the reason why )

starter-code/css/main.css
```diff
@@ -6,3 +6,110 @@ Green: #00B172
...
+    font-size: 2em;
+}
+
+/* whats-on */
+
+.whats-on {
+    margin: 100px 20px 230px 0px;
+}
+
+.whats-on h2 {
```
ianizaguirre: I will admit that  I think this part was hard to get right 😁, you would actually not use a bottom border green ....

You could have done something like this for the HTML:

```html
                    <div class="fancy-border-green">
                        <h2>What’s on Spotify?</h2>
                    </div>
```

and this for the CSS:
```css
.fancy-border h2 {
    text-align: center;
}
.fancy-border-green {
	width: 360px;
	position: relative;
	margin: 0 auto;

}
.fancy-border-green:after {
    border-bottom: 4px solid #1db954;
    content: '';
    height: 1px;
    position: absolute;
    left: 0;
    bottom: -5px;
    width: 100%;
}
```

ianizaguirre:
You did good, even though I know you where not finished working on it. Keep improving, because you are getting better !

Kind Regards,
Ian Izaguirre | Ironhack TA

PR #432, opened 2018-04-05
[MIA]-FTWD-RANDY-ARBOLAEZ-LABS-CSS-SPOTIFY-CLONE

randyarbolaez:


starter-code/index.html
```diff
@@ -1,42 +1,103 @@
...
+  <div>
+    <section>
+      <h2 class="what-spotify">What’s on Spotify?</h2>
+      <hr class="green-line">
+    </section>
+
+    <section class="middle">
+      <div class="box">
+        <img class="middle-images" src="./images/music-icon.png" alt="music-icon">
+        <section class="text">
```
sandrabosk: I would change all these inside section tags with div tags, just because section tag groups part of a page that logically belongs together so making wrapper with section tag makes sense but having a lot of inside section tags really doesn't. Check this:
https://stackoverflow.com/questions/6939864/what-is-the-difference-between-section-and-div

sandrabosk:
Great job! Thank you for submitting.

PR #433, opened 2018-04-05
FTWD Miami - Ryan Martin

ryancraigmartin:


starter-code/css/main.css
```diff
@@ -1,8 +1,141 @@
...
+
+}
+
+.millions-of-songs{
+
+}
+
+.music-icon{
+  height: 100px;
+  width: 100px;
```
MarcosPalaciosIII: if you have multiple images or tags in general that use the same styling, you can just use the same class name for all the tags you want to target and it helps to keep your css clean so as to not repeat code over multiple times.

starter-code/index.html
```diff
@@ -2,41 +2,76 @@
...
+  <h1 class="main-heading">Music for everyone.</h1>
+  <h2 class="sub-heading">Spotify is now free on mobile, tablet and computer.</h2>
+  <h2 class="sub-heading">Listen to the right music, wherever you are.</h2>
+</div>
+
+<div class="center-section">
+
+  <h3 class="center-section-heading">What’s on Spotify?</h3>
+
+    <div class="center-section-group"></div>
```
MarcosPalaciosIII: bottom 3 divs would go into this div if you are planning on modifying the 3 together for positioning.

MarcosPalaciosIII:
overall good work. I'm sure with a little more time you would have been able to finish this.  Everything on the top portion looks really good. Maybe just a bit of adjustment on the banner background image.

PR #438, opened 2018-04-05
MIA-FTWD-Kat Alderman - lab-css-spotify-clone

katalderman:
first commit, but i'm not done yet. to be completed 4/5

### DIFF
--- a/starter-code/css/main.css
+++ b/starter-code/css/main.css
@@ -64,6 +64,12 @@ nav {
   padding-top: 10px;
 }
 
+nav a{
+  color: black;
+  text-decoration: none;
+  font-size: 24px;
+}
+
 .navigation {
   float: right;
   padding: 0px 10px;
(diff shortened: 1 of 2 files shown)


starter-code/css/main.css
```diff
@@ -6,3 +6,187 @@ Green: #00B172
...
+/* image styles */
+
+.logo {
+  max-width: 220px;
+  max-height: 70px;
+}
+
+/* structure styles */
+
+nav {
```
sandrabosk: Let's add class to "nav" next time so you don't get the habit to use tags since with Bootstrap it will get messy (you will overwrite the predefined Bootstrap classes if you keep doing this).

starter-code/index.html
```diff
@@ -3,40 +3,94 @@
...
+      smartphone, tablet or
+      computer</p>
+    </div>
+
+
+  <div class="clear"></div>
+</section>
+
+
+<section id="greenBackground"></section>
```
sandrabosk: This 
```
</section>
```
here is extra.

sandrabosk:
Amazing work Kat. I like that you used that many classes. Thank you for submitting.

PR #440, opened 2018-04-05
ongoing update

olaborde:


MarcosPalaciosIII:
remember to check out the links when cloning a site. it usually has more info that you can use. Also try using padding and margin some more to help adjust positioning. When cloning detail is important, try and match up as much of what your cloning as possible like order of lists and/or links.

overall good work. just need to finish the bottom half of the clone. https://s3-eu-west-1.amazonaws.com/ih-materials/uploads/spotify-prototype.pdf you can go here to review the rest of the clone or enter your lab and click the link under the example.

PR #442, opened 2018-04-06
MIA - FTWD - RAFAEL GUASP - Spotify Clone

rguasp:
not done need to tweak it around.Will submit new request on weeknd.

### DIFF
--- a/starter-code/css/main.css
+++ b/starter-code/css/main.css
@@ -6,3 +6,131 @@ Green: #00B172
 White: #FFF
 
 */
+
+body{
+  margin: 0 auto;
+}
+.spotifylogo{
+width: 150px;
+height: 50px;
+padding-left: 30px;
+padding-top: 12px;
+
+}
+.top-right li{
+  
+  display: inline-block;
+  margin: 0 10px;
+  
+  
+}
+nav{
+height: 75px;
+margin: auto 0;
+background-color: white;
+position: fixed;
+width: 100%;
+
+}
+.top-right{
+  float: right;
+  padding-right: 30px;
+  padding-top: 12px;
+
+}
+nav a{
+  text-decoration: none;
+  color: black;
+}
+.container{
+
+  height: 480px;
+  width: 100%;
+  background-image:url("../images/landing.jpg");
+  background-size: contain;
+  
+}
+.containertitle{
+  color: white;
+  width: 670px;
+  font-size: 2rem;
+  padding-top: 20vh;
+  padding-left: 25vh;
+  text-align: center;
+
+}
+.containertitle p{
+  line-height: 0;
+}
+.container2 h2{
+  padding-left: 380px;
+  font-size: 2rem;
+  margin-bottom: -5px;
+
+}
+.container2head hr{
+  height: 3px;
+  background-color:#00cc99;
+  width: 250px;
+  border: 0;
+ 
+
+}
+.container3 img{
+  width: 100px;
+  height: 100px;
+}
+.container3 h3{
+  color:#00cc99;
+  font-size: 1.5rem;
+}
+.MillofSong {
+  float: left;
+  width: 33%;
+
+}
+.Hdmusic{
+  float: left;
+  width: 33%;
+}
+.StreamEverywere{
+  float: left;
+  width: 33%;
+
+}
+.container3{
+height: 500px;
+padding-left: 150px;
+padding-top: 100px;
+}
+.kanye-west-wrapp{
+  background-color: #00B172;
+  margin: auto 2% 2% 2%;
+}
+
+.kanye-west h2{
+  display: inline;
+  color: #FFF;
+  font-size: 2.2em;
+  border-bottom: 2px solid #FFF;
+  border-bottom-width: medium;
+  padding-bottom: 5px;
+  font-weight: lighter;
+}
+
+.kanye-west h3{
+  font-size: 2em;
+  color: #FFF;
+}
+
+.kanye-west p{
+  color:#FFF;
+  font-size: 1.5em;
+  font-weight: lighter;
+}
+
+.screen-capt{
+  height: 20%;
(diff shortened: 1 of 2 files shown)


starter-code/index.html
```diff
@@ -3,40 +3,81 @@
...
+</div>
+
+<div class="Hdmusic">
+<img src="./images/high-quality-icon.png" alt="Hdmusic">
+<h3>HD Music</h3>
+<p>Listen to music as if you</p>
+<p>were listening live</p>
+</div>
+
+<div class="StreamEverywere">
```
ianizaguirre: Throughout this entire HTML page, you decided to name SOME of your class names using camel case? Please only use camel case in Javascript. The convention everyone follows for naming classes in HTML is to use all lower case and separate words with a hyphen.  So for example, your `"StreamEverywere"` should have been `"stream-everywere"` ... A few reasons why everyone follows this rule, is because CSS syntax is mostly case-insensitive. So to avoid errors others can have working on a team project with your code or reading your code, its best to just keep everything lower case and to separate words with a hyphen. It's best that you just realize that the best approach is the most compatible: use the same case in all code for a given CSS class, and don't create two or more CSS class names that differ only in case. It is also a good practice for readability. Also, even though there are other accepted ways to name classes in HTML, like using an underscore instead of a hyphen, for now lets stick with a hyphen.

ianizaguirre: Also I think you misspelled this part of your class name:  "everywere"

starter-code/index.html
```diff
@@ -3,40 +3,81 @@
...
+</div>
+</div>
+
+<div class="container2">
+  <div class="container2head">
+<h2>What’s on Spotify?</h2>
+<hr>
+</div>
+</div>
+<div class="container3">
```
ianizaguirre: Please start getting into the habit of properly spacing/tabbing ALL of your divs, because all of your divs are on the same left side as the body tag, so now it is hard to follow where starting and closing tags start and end.

starter-code/index.html
```diff
@@ -3,40 +3,81 @@
...
+</nav>
+<div class="container">
+  <div class="containertitle">
+<h1>Music for everyone.</h1>
+<p>Spotify is now free on mobile, tablet and computer.</p>
+<p>Listen to the right music, wherever you are.</p>
+</div>
+</div>
+
+<div class="container2">
```
ianizaguirre: Calling things `"container1"` , `"container2"`  , `"container3"`  etc.  should be avoided. Names like `"container1"`  are way to general. Choose a better name since when we make div classes the plan is to reuse them eventually in our code in other places...So if the name is descriptive, it wont look out of place in another part of our HTML code when we reuse it, and a good name can tell us information.
For example:
 `  <div class="container"> `
could have been better written as something like:
 `  <div class="hero-wrapper"> `
or
 `  <div class="hero-container"> `

starter-code/css/main.css
```diff
@@ -6,3 +6,131 @@ Green: #00B172
 White: #FFF
 
 */
+
+body{
+  margin: 0 auto;
```
ianizaguirre: It would be a good habit to include a width in our body, so add: `width: 100%;`

starter-code/css/main.css
```diff
@@ -6,3 +6,131 @@ Green: #00B172
...
+
+}
+.top-right li{
+  
+  display: inline-block;
+  margin: 0 10px;
+  
+  
+}
+nav{
```
ianizaguirre: Properly space/tab your CSS rules so the properties are nested under the rule.

starter-code/index.html
```diff
@@ -3,40 +3,81 @@
...
+<p>smartphone, tablet or</p>
+<p>computer</p>
+</div>
+</div>
+
+<div class="kanye-west">
+  <div class="kanye-west-wrapp">
+    <h2>It’s as yeezy as Kanye West.</h2>
+
+    <div class="search-browse-discover">
```
ianizaguirre: "search-browse-discover" is a little too long of a class name, try to make it smaller

starter-code/index.html
```diff
@@ -3,40 +3,81 @@
...
+<img src="./images/devices-icon.png" alt="StreamEverywere">
+<h3>Stream Everywhere</h3>
+<p>Stream music on your</p>
+<p>smartphone, tablet or</p>
+<p>computer</p>
+</div>
+</div>
+
+<div class="kanye-west">
+  <div class="kanye-west-wrapp">
```
ianizaguirre: `<div class="kanye-west-wrapp">`, 
I think you either mean to write `<div class="kanye-west-wrap">` or `<div class="kanye-west-wrapper">`

starter-code/index.html
```diff
@@ -3,40 +3,81 @@
...
-Just search and hit play.
-Browse
-Check out the latest charts,
-brand new releases and great
-playlists for right now.
-Discover
-Enjoy new music every Monday
-with your own personal playlist.
-Or sit back and enjoy Radio.
+    <nav>
```
ianizaguirre: You are not writing any semantic tags in your HTML code? You are forgetting to wrap your nav with header tags. The `<header>` element is used to contain the sites name and the main navigation. 
You are also forgetting to include `<section>` tags. The `<section>` element groups related content together, typically each section would have its own heading.

ianizaguirre:
I know you said in your comment, that you are still working on this. So I just left comments on what I have seen so far. I will leave this pull request OPEN since you are still working on it. When you are done working on it, Slack me a message so that I can check it for you.

Try to apply this feedback on this current project and your next project.

Kind Regards,
Ian Izaguirre | Ironhack TA

rguasp:
i tried my best to improve on the mistakes you pointed out. For some reason i was having trouble using flexbox for the text and image alignment I know im not supposed to use margin but was out of ideas.

PR #443, opened 2018-04-06
MIA - Claudia S. - spotify clone

clausan121:


MarcosPalaciosIII:
seems you were starting to head in the right direction. Talk to me so we can go over the clone.

PR #480, opened 2018-05-23
BER - Silvio

silvio-galli:
The bounce effect is an adaptation from a codepen.


ta-web-ber:
Very good code :dancer: 
I tried to improve the responsiveness.
But without using flex and/or Bootstrap it's too messy :)

If you want rock css animations, take a look on [it](http://animista.net/)
_Vivian_

PR #483, opened 2018-05-23
BER - Jens

je0705:
Done Version 1

### DIFF
--- a/starter-code/css/main.css
+++ b/starter-code/css/main.css
@@ -4,5 +4,124 @@ Colors:
 Text: 1A1A1A
 Green: #00B172
 White: #FFF
-
 */
+* {
+  font-family: Arial, Helvetica, sans-serif;
+}
+
+.nav-container {
+  background-color: #FFF;
+  height: 12vh;
+  width: 100wh  ;
+
+}
+.nav-bar-left > img {
+  height: 80px;
+  width: 300px;
+  float: left;
+  margin: 5px;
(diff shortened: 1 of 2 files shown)


starter-code/css/main.css
```diff
@@ -4,5 +4,124 @@ Colors:
...
+  margin-bottom: 60px;
+
+}
+  
+h3 {
+  color: #00B172;
+}
+
+.info-box > img {
+  height: 20vh;
```
ta-web-ber: Change only the width in order to respect the homothety of the image.
By the way, the css selector is really neat, I like it!

starter-code/css/main.css
```diff
@@ -4,5 +4,124 @@ Colors:
...
+  width: 20vw;
+}
+
+.green-container {
```
ta-web-ber: :+1:

starter-code/css/main.css
```diff
@@ -4,5 +4,124 @@ Colors:
...
+  text-align: left;
+  color: #FFF;
+
+  background-color: #00B172;
+  background-position: center;
+  background-image: url('../images/spotify-icon-white.png');
+  background-repeat: no-repeat;
+  background-size: 10%;
+}
+.green-box {
```
ta-web-ber: You should create left and right classes.
Because the Kayne West asset doesn't have the same properties.

PR #486, opened 2018-05-24
BER - TING

tjHsu:
Just want to ask, if we want to make the spotifyLogo jumps up and down, which one is better, transition or animation?

### DIFF
--- a/starter-code/css/main.css
+++ b/starter-code/css/main.css
@@ -19,19 +19,19 @@ body{
 #nav img{
     width: 100px;
     float: left;
-    margin-bottom:10px;
+    /* margin-bottom:10px; */
 }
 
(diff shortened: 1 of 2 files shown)


ta-web-ber:
I prefer CSS animations.
You could find good ones here:
http://animista.net/

starter-code/css/main.css
```diff
@@ -6,3 +6,125 @@ Green: #00B172
...
+#footer{
+    clear: both;
+    color: white;
+    text-align: left;
+}
+
+
+
+#footer div{
+
```
ta-web-ber: Add 
```css
padding-top: 10vh;
```
It makes the div cleaner :)

PR #506, opened 2018-06-06
PAR diane

Dianedel:
I would like finish this exercise before this week-end.


ta-web-paris:
Ok I'm letting the pull request open, will take a look a it on monday

PR #524, opened 2018-06-13
MIA/AlexZarrella/Spotify-clone

AlexZarrella:


MarcosPalaciosIII:
excellent work on the spotify clone. Just needed to make the nav links a little bigger.  

Keep up the excellent work

PR #534, opened 2018-06-14
MIA-Andrew Flory

DrewFlory:


starter-code/css/main.css
```diff
@@ -6,3 +6,158 @@ Green: #00B172
...
+}
+
+.nav li {
+    padding-right: 20px;
+}
+
+.top-section {
+    background-size: cover;
+    background-repeat: no-repeat;
+    background-image: url('/Users/Drewskii/Desktop/labs/code/lab-css-spotify-clone/starter-code/images/landing.jpg')
```
MarcosPalaciosIII: your path should be `background-image: url('../images/landing.jpg')`. websites can't read a path that is on your pc unless its from your pc. the `../` means go out 1 folder, then you do `images/` to look for the images folder and then you choose the file `landing.jpg`.  if you don't have to leave the current folder your in like in your index html, then you use `./` which is to stay in current folder.

starter-code/css/main.css
```diff
@@ -6,3 +6,158 @@ Green: #00B172
...
+
+* {
+    font-family: 'Raleway', sans-serif;
+}
+
+.nav {
+    background-color: white;
+    width: 100vw;
+    height: 60px;
+    top: 12px;
```
MarcosPalaciosIII: `top: 0` sticks it to the top of the page. otherwise you notice the background just above the nav bar as you scroll.

starter-code/css/main.css
```diff
@@ -6,3 +6,158 @@ Green: #00B172
...
 */
+
+* {
+    font-family: 'Raleway', sans-serif;
+}
+
+.nav {
+    background-color: white;
+    width: 100vw;
+    height: 60px;
```
MarcosPalaciosIII: nav needs a bit more height. 80px would have been good for this one.

starter-code/css/main.css
```diff
@@ -6,3 +6,158 @@ Green: #00B172
...
+    top: 12px;
+    position: fixed;
+}
+.nav img {
+    margin-top: -4px;
```
MarcosPalaciosIII: unless you want something to go off the screen you usually would not add negative margin.

starter-code/css/main.css
```diff
@@ -6,3 +6,158 @@ Green: #00B172
...
+    height: 50px;
```
MarcosPalaciosIII: when adjusting image sizes, you usually only manipulate the width or the height. Controlling the width auto adjust the height and vise versa, Unless you need a specific height and width but many times that distorts the image from what it should look like.

starter-code/css/main.css
```diff
@@ -6,3 +6,158 @@ Green: #00B172
...
+}
+
+
+.text {
+   position: relative;
+   text-align: center;
+    top: -420px;
+    font-size: 23px;
+    color: white;
+    width: 26%;
```
MarcosPalaciosIII: text would have fit more like the clone with a width of 38% rather than 26%.

starter-code/css/main.css
```diff
@@ -6,3 +6,158 @@ Green: #00B172
...
+    display: flex;
+    flex-direction: row;
+    margin-top: 70px;
+    justify-content: center;
+}
+
+.million, .hd, .stream {
+    display: flex;
+    flex-direction: column;
+    padding-right: 50px;
```
MarcosPalaciosIII: a little more padding on both sides. about 130px would have been good for the spacing. Also another way to control the padding on both sides is like this `padding: 0 130px 0 130px`. the first number is for the top padding then the right, bottom and finally left.  It goes clockwise.

MarcosPalaciosIII:
overall good work on the spotify clone. Just keep in mind when cloning to keep the page your cloning open to have a good reference as to how things should be styled or positioned. 

keep up the good work.

PR #552, opened 2018-06-14
MIA061118_FT_Alina_K

akitaeva:
final draft
-margin on the left to be fixed
-fonts adjusted


MarcosPalaciosIII:
Excellent work on the spotify clone. The bottom section's text seems a bit small though. Remember when cloning its good to have the page your cloning open so that it is easier to compare your work to what your cloning. 

keep up the excellent work.

PR #620, opened 2018-08-02
spotify, not finished

tlebon:
didnt take enough time to finish unfortunately.


ta-web-ber:
Good job so far. It was a tricky exercise. You missed a cpuple of details but it is totally normal.

When you create the pull request, you should give it a title like this:
#BER - Timothy
so we can find it easily among the pull requests from other campuses.

thanks
Silvio

PR #639, opened 2018-08-14
PAR Pauline & Mohamed

SMoZam:
@paulettepaillette


ta-web-paris:
This looks really nice !! You'll have to give me css lessons ;)
one small comment : remember to fill in the alt tags: even if they don't change anything visually, they are absolutely necessary for several reasons, and accessiblity is in itself a good enough reason !

PR #640, opened 2018-08-14
par Constance & Nerimene

nerimeneakari:
@Cr27910

### DIFF
--- a/starter-code/css/main.css
+++ b/starter-code/css/main.css
@@ -6,3 +6,158 @@ Green: #00B172
 White: #FFF
 
 */
+html{
+    font-family: 'Montserrat', sans-serif;
+}
+
+body{
+    margin: auto;
+    padding: 0;
+}
+
+header{
+    display: flex;
+    justify-content: space-between;
+    background-color:white;
+}
+
+.logo {
+    width: auto;
+    height: 50px;
+    margin: 20px;
+}
+
+nav {
+    display: flex;
+    justify-content: space-around;
+    margin: 20px;
+}
+
+ul {
+    display: flex;
+    justify-content: space-between;
+}
+
+a{
+    text-decoration: none;
+    color: #1A1A1A;
+    margin: 0 10px;
+    font-size: 16px;
+    font-weight: 300;
+}
+
+li{
+    display: flex;
+    justify-content: space-around;
+    list-style: none;
+}
+
+#articles{
+    display: flex;
+    justify-content: space-around;
+}
+
+#articles article{
+    text-align: center;
+    margin: 20px 0;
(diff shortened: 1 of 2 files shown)


starter-code/css/main.css
```diff
@@ -6,3 +6,158 @@ Green: #00B172
...
+    color: white;
+    display: flex;
+    flex-direction: column;
+    align-items: center;
+    justify-content: center;
+    background-image: url("../images/landing.jpg");
+    background-position: 50% 50%;
+    background-repeat: no-repeat;
+    background-size: cover;
+    height: 600px;
```
ta-web-paris: good use of the backgound properties!
To see more of the image, you can increase the height to 1000px for example

ta-web-paris:
Your CSS code is efficient, you are using the right properties in the correct way.
To adjust the spacing in the spotlight section, you can rearrange a litlle bit your html in that section to have a wrapper div then set it a width (for example 80%), and make it a flex box with justify-content : space-between.
inside this wrapper div, there would be two divs, one for the text and one for the image, and you can also make a flexbox out of the text div to deal with the spacing between paragraphs.

starter-code/index.html
```diff
@@ -1,42 +1,106 @@
...
+    <br>Listen to the right music, wherever you are.</p>
+    
+  </section>
+
+  <section >
+
+    <h2><span>What’s on Spotify?</span></h2>
+
+    <div id="articles">
+      <article>
```
ta-web-paris: Usually, the article tag is used for content that is relatively independant from the rest of the website page or that yould be used outside of the context of this page, I would rather use divs in this case.

ta-web-paris: more info : http://html5doctor.com/the-article-element/

starter-code/index.html
```diff
@@ -1,42 +1,106 @@
...
+
+        <img src="./images/music-icon.png" alt="music">
+        <h3>Millions of Songs</h3>
+        <p>There are millions of
+        <br>songs on Spotify</p>
```
ta-web-paris: <br> tags also have a specific semantic meaning, they should only be used when the line break does not serve styling purpose but is part of language conventions, like in poems of addresses.
you can either put : 
```
        <img src="./images/music-icon.png" alt="music">
        <h3>Millions of Songs</h3>
        <p>There are millions of</p>
        <p>songs on Spotify</p>
```
or 
```
        <img src="./images/music-icon.png" alt="music">
        <h3>Millions of Songs</h3>
        <p>There are millions of
        <span>songs on Spotify</span></p>
```
and add a dispaly: block to the span in the css.

PR #641, opened 2018-08-14
PAR Paul & Alice

alicedbrg:


starter-code/index.html
```diff
@@ -2,41 +2,79 @@
...
+
+<section class="features-title">
+  <h2>What’s on Spotify?</h2>
+
+<div class="features">
+  <div class="features-desc">
+    <img src="./images/music-icon.png" alt="music-icon">
+    <h3>Millions of Songs</h3>
+    <p>There are millions of<br>
+    songs on Spotify</p>
```
ta-web-paris: br tags have a specific semantic meaning, they should only be used when the line break does not serve styling purpose but is part of language conventions, like in poems of addresses.
you can either put :
```
        <img src="./images/music-icon.png" alt="music">
        <h3>Millions of Songs</h3>
        <p>There are millions of</p>
        <p>songs on Spotify</p>
```
or better:
```
        <img src="./images/music-icon.png" alt="music">
        <h3>Millions of Songs</h3>
        <p>There are millions of
        <span>songs on Spotify</span></p>
```
and add a dispaly: block to the span in the css.

Or even better : just have one p tag and use the flex-wrap property to deal with the line break.

starter-code/css/main.css
```diff
@@ -6,3 +6,160 @@ Green: #00B172
...
+    padding-top: 100px ; 
+    color: white;
+    margin-bottom: 50px;
+}
+
+.title-img h1 {
+    text-align: center;
+    font-size: 65px; 
+    letter-spacing: 3px;
+    margin-top: 300px;
```
ta-web-paris: This is the property that was messing with your flow and adding a weird white space above when we checked the code together. If you replace it with a padding-top, then you can remove the top: 0 property from the header.

ta-web-paris:
Good job on the layout!  It is all a matter of spacing now. 
There are tools to identify what font a website is using, this is the one I use : https://chrome.google.com/webstore/detail/whatfont/jabopobgcpjmedljpbcaablpmlmfcogm?hl=en

PR #642, opened 2018-08-15
PAR Manon & Emmanuelle

emmanuellejaldorau:
@msln8

### DIFF
--- a/starter-code/index.html
+++ b/starter-code/index.html
@@ -3,40 +3,58 @@
 <head>
   <meta charset="UTF-8">
   <title>Spotify</title>
+  <link rel="stylesheet" href="./css/main.css">
 </head>
 <body>
-Premium Discover Help Download
+  <header>
+    <img src="./images/spotify-logo.png" alt="">
+    <nav>
+      <a href="#0">Premium</a>
+      <a href="#0">Discover</a>
+      <a href="#0">Help</a>
+      <a href="#0">Download</a>
+    </nav>
+  </header>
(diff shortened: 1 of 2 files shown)


starter-code/index.html
```diff
@@ -3,40 +3,58 @@
...
+      </div>
+      <div>
+          <img src="./images/high-quality-icon.png" alt="">
+          <h3>HD Music</h3>
+          <p>Listen to music as if you<br/> were listening live</p>
+        </div>
+        <div>
+            <img src="./images/devices-icon.png" alt="">
+            <h3>Stream Everywhere</h3>
+            <p>Stream music on your<br/> smartphone, tablet or<br/> computer</p>
```
ta-web-paris: br tags should only be used for poems, addresses, or other cases where the line break is part of language conventions, line break for styling purposes are better dealt with from the css. 
By using flexbox and managing the width of your box, you should be able to have the line break where you intend to (you may have to use the flex-wrap property). If it doesnt end up being like you wanted it to be, here is another way : 
```
<div>
            <img src="./images/devices-icon.png" alt="">
            <h3>Stream Everywhere</h3>
            <p>Stream music on your
            <span> smartphone, tablet or</span>
            <span>computer</span></p>
</div>
```
then you can put a display: block to your span in the css. This technique is useful for two-lines titles.

ta-web-paris:
Nice job ! I like your use of the parent selectors in the css. :100:

PR #643, opened 2018-08-15
PAR Aurélien & Luca

lucamerzi:
@LOOFFI


starter-code/index.html
```diff
@@ -3,40 +3,100 @@
...
+   
+</header>
+
+<!-- First section -->
+
+<section>
+	<h1>Music for everyone.</h1>
+	<h4>
+		Spotify is now free on mobile, tablet and computer.
+		<br>
```
ta-web-paris: Like I said to others, br tags should only be used for poems, addresses, or other cases where the line break is part of language conventions; line break for styling purposes are better dealt with from the css.
By using flexbox and managing the width of your box, you should be able to have the line break where you intend to (you may have to use the flex-wrap property). You can also wrap the second line in a span and add a display: block to this span in the css!

ta-web-paris:
Your page looks really good, good job ! :100:

PR #645, opened 2018-08-15
MIA FTWD MikeWIthers

mwithers313:


MarcosPalaciosIII:
good work on  the spotify clone. Remember that you don't always have to put width on div's unless you want to make them smaller than their size or changing display: block in any way as display: block usually makes the div the width of the page. 

Keep up the good work.

PR #648, opened 2018-08-15
MIA WebDev FT: Sat

satga:
can't get the sizing to work on the bottom green section when changing window size

### DIFF
--- a/starter-code/css/main.css
+++ b/starter-code/css/main.css
@@ -6,3 +6,137 @@ Green: #00B172
 White: #FFF
 
 */
+html {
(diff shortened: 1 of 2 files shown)


MarcosPalaciosIII:
good work on the spotify clone. For the main image you can use background-position: center to have the center of the image be the main part of the image that shows. For the section below Whats On Spotify, you could put the 3 articles in a div to adjust the positioning better within the div and then give the div any padding or margin needed to have everything centered. The bottom green section has the logo a little low and the image a is bit large which makes it go below the green background. 

Keep up the good work.

PR #650, opened 2018-08-15
MIA-FTWD-Tony Batuhan Dogan

batuhaandogan:
Couldn't finish the Navbar.


MarcosPalaciosIII:
nice work on the spotify clone. your bottom div is a little to tall and the text should be moved over more. The nave bar would need the links to have no style and a little larger. For the div w/ the main background, you have a little to much line height. 

keep up the nice work.

PR #651, opened 2018-08-16
MIA FTWD Jonathan Harris

yonathanh:


MarcosPalaciosIII:
great work on the spotify clone. For the nav bar the links should be moved over a little closer to the edge of the screen. The green section on the bottom has a little too much height.

Keep up the great work.

PR #652, opened 2018-08-16
PAR Abi 

abisummers:


starter-code/index.html
```diff
@@ -1,42 +1,84 @@
...
+    </div>
+  </section>
+
+  <section>
+
+    <div class="green-box">
+
+      <div class="left">
+        <h2>It’s as yeezy as Kanye West.</h2>
+        <div>
```
ta-web-paris: You can set a width to this div to have the line breaks where you want !

ta-web-paris:
good job!! your code is well organised and the result is great :100:

PR #653, opened 2018-08-16
spotify-Levijs-Mia-FTWD

levijskal:


MarcosPalaciosIII:
nice attempt on the spotify clone. Though your css seems to be off, basically all of your images and text are really large. I tried different screen sizes but nothing lines up. 

this one will need some improvement.

PR #655, opened 2018-08-16
MIA FTWD - Spotify Clone - Lauren Cahill

laurencahill:


MarcosPalaciosIII:
great work on the spotify clone. The background image under the nav should be stretched to cover the div though. 

Keep up the great work.

PR #657, opened 2018-08-20
MIA ftwd alex rojas

alexrojas:


MarcosPalaciosIII:
good work on the spotify clone. For the nav bar the links seem to be aligned off center. The main background image is a be small, the div should have some more height. the 2nd div content has the font a little small and the images with the text should be more centered on within each div. For the final section the content is a little small. the div itself would need more height and the text along with the images should be larger. 

keep up the good work.

PR #658, opened 2018-08-21
Inna MIA FTWD

innayugay:
partially done


MarcosPalaciosIII:
good attempt on the spotify clone. You seemed to be on the right track as far as styling. The main take away is that when styling, probably one of the main things you wan't to focus on first is positioning. The actual detailing is easier after that.

Keep up the good work.

PR #660, opened 2018-08-22
[MAD de] Francis Wardley

francisgeorge90:
I found the HTML and CSS relatively easy, but the animation was tricky so I used https://daneden.github.io/animate.css/ to help.


ta-web-mad:
Sigue así, está genial!
Anna

PR #727, opened 2018-09-27
[PAR-PT] ArthurHoudebine

arthurhoudebine:
NavBar 75% done
Header image : cannot align titles
@ta-webpt-par

### DIFF
--- a/starter-code/css/main.css
+++ b/starter-code/css/main.css
@@ -6,3 +6,92 @@ Green: #00B172
 White: #FFF
 
 */
+
+* {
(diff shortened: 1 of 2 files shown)


starter-code/css/main.css
```diff
@@ -6,3 +6,92 @@ Green: #00B172
...
 */
+
+* {
+font-family: Circular, Arial, Helvetica, sans-serif;
+
+}
+
+.navbar{
+background-color:white;
+position : fixed;
```
ta-webpt-par: You don't have to fix it because it will tend to block the position even when you scroll down.

PR #728, opened 2018-09-27
[PAR-PT]Elodie

EloK6:
Just the beginning
@ta-webpt-par

### DIFF
--- a/starter-code/css/main.css
+++ b/starter-code/css/main.css
@@ -6,3 +6,47 @@ Green: #00B172
 White: #FFF
 
 */
+
+body {
+    margin: 0 auto;
+}
+
+.header {
+    height: 100px;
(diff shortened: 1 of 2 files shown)


starter-code/css/main.css
```diff
@@ -5,48 +5,174 @@ Text: 1A1A1A
...
    height: 50px;
    float: left;
+   margin: 20px 0 0 20px;
 }
 
 .menu-nav {
     list-style-type: none;
     text-align: center;
-    margin: 0 auto;
+    margin: 0 20px;
```
ta-webpt-par: You can get rid of the margin and only use the padding and the flexbox 
   display : flex;
and 
  padding: 30px 50px 0 750px;

PR #729, opened 2018-09-28
<PAR PT> Sandra

SanDDupt:
"1ère version"
@ta-webpt-par

### DIFF
--- a/starter-code/css/main.css
+++ b/starter-code/css/main.css
@@ -1,4 +1,4 @@
-/*
+  /*
 Colors:
 
 Text: 1A1A1A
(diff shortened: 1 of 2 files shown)


starter-code/css/main.css
```diff
@@ -6,3 +6,161 @@ Green: #00B172
...
+body {
+  width: 100%;
+  padding: 10px;
+  text-align: center;
+}
+
+header {
+  width: 100%;
+  height: 100px;
+  position: fixed;
```
ta-webpt-par: I would have gotten rid of the
        position : fixed
because it makes the header appear when you scroll down and would have added 
        background-color: #00B172;
        color: white;

starter-code/css/main.css
```diff
@@ -6,3 +6,161 @@ Green: #00B172
...
+  transition-property: margin-left;
+  transition-duration: 1.5s;
+  transition-timing-function: linear;
+  transition-delay: 0.3s;
+}
+header .navBar {
+  float: right;
+  width: 60%;
+  vertical-align: middle;
+  margin-top: 38px;
```
ta-webpt-par: It would be better to use the padding and maybe the flexbox

SanDDupt:
Merci !

Le lun. 1 oct. 2018 à 16:18, ta-webpt-par <notifications@github.com> a
écrit :

PR #759, opened 2018-10-18
Berlin-Rebeca-Cabañas-WDPT2018

beckyarauz:


NicholsonJ:
Great job on the exercise.  I love the carousel! lol 

If you want to adjust this exercise, you can correct the following problem. It only happens when the screen size is reduced.  

screen shot 2018-10-23 at 7 27 16 pm

beckyarauz:
awesome! I'll work on it! :D thanks a lot!

PR #762, opened 2018-10-18
Liska-Crofts

LiskaC:
first edits


NicholsonJ:
Good start to the exercise. The size/positioning of the header element and the background/cover image look good.  Feel free to come back to this exercise and continue working for more css practice.  You can also style the links in the nav bar, which will make them look a little nicer.

PR #763, opened 2018-10-18
Berlin-Lara-Petersen

withenoto:


NicholsonJ:
Great start to the exercise! The nav bar and cover image look great.  Your centering on the text looks nice as well. Try adding a class to the div that contains the iconbox elements.  Then you can style them in relation to a larger element. Good start!

NicholsonJ:
Great start to the exercise! The nav bar and cover image look great. Your centering on the text looks nice as well. Try adding a class to the div that contains the iconbox elements. Then you can style them in relation to a larger element.

PR #764, opened 2018-10-18
Berlin-Sarah-Spotify

sarahdrs:


NicholsonJ:
Good start! Make sure the link to the style sheet happens in the "<head>" section.  Your nav bar looks great and the size of the background cover image looks good.  Feel free to come back to this exercise to play around with css.

PR #767, opened 2018-10-19
BER-Steph-Gidde-done

StephGidde:


NicholsonJ:
Great work! Your styling follows the example really well! Things to play with (if you wanted) would be to try making the page more responsive.  But well done.

PR #768, opened 2018-10-19
Berlin - Susanne Tünker - practically finished

su-jaya:
will improve my coding during the next week


NicholsonJ:
Great work on the exercise! Beneath the "whatsOnSpotify" section, create a new div to hold all of the containers.  That way you can style them relative to a larger container, instead of estimating sizes for each object.

PR #776, opened 2018-10-22
Paris - Amine

enima2684:


starter-code/css/style.scss
```diff
@@ -0,0 +1,54 @@
+body{
+  margin: 0;
```
ta-web-paris: Well Done Amine!!

everything seems to look fine. THere is just a very small alignment problem on the 3rd block just below the header. This is caused by the 2 lines of text (instead of 1 like the others) but that's a very small fix to do.

PR #777, opened 2018-10-22
done -Paris  Niccolò

nicrodemus:
i think the spotify logo in the green box can be upgraded.


nicrodemus:
the second push is with the logo animated

PR #780, opened 2018-10-22
Mathis Spotify Clone

Mathispnz:
Struggle to find clear and logic names for the tags !!

### DIFF
--- a/starter-code/css/main.css
+++ b/starter-code/css/main.css
@@ -6,3 +6,171 @@ Green: #00B172
 White: #FFF
 
 */
+
+body {
+  margin: 0;
+  font-family: -apple-system, BlinkMacSystemFont, 'Segoe UI', Roboto, Oxygen, Ubuntu, Cantarell, 'Open Sans', 'Helvetica Neue', sans-serif;
+  font-weight: 300;
+}
+
+/* HEADER */
+
+header {
+  margin: 20px auto;
+  background-color: white;
+  height: 30px; 
+  width: 95%;
+  display: flex;
+  flex-direction: row;
+  justify-content: space-between;
+  align-items: center;
+}
+
+header img {
+  width: 170px;
+  height: 50px;
+}
+
+header ul {
+  display: flex;
+  flex-direction: row;
+  justify-content: flex-end;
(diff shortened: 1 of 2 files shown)


starter-code/css/main.css
```diff
@@ -6,3 +6,171 @@ Green: #00B172
...
+header li {
+  padding: 10px 20px;
+}
+
+header a {
+  color: black;
+  text-decoration: none;
+}
+
+@keyframes jump {
```
ta-web-paris: cool animation!

starter-code/css/main.css
```diff
@@ -6,3 +6,171 @@ Green: #00B172
...
+body {
+  margin: 0;
+  font-family: -apple-system, BlinkMacSystemFont, 'Segoe UI', Roboto, Oxygen, Ubuntu, Cantarell, 'Open Sans', 'Helvetica Neue', sans-serif;
+  font-weight: 300;
+}
+
+/* HEADER */
+
+header {
+  margin: 20px auto;
```
ta-web-paris: by adding 

```
header {
position: fixed;
}
```

your header would remain in place when you scroll. You'd also need to change the with to 100%. 

it's also better to increase the height of the header rather than adding a margin. Adding margin can be useful sometimes, but can also interfere with other css styling later. So here you could just set the height to 60px

starter-code/index.html
```diff
@@ -3,40 +3,83 @@
...
+  <p>Spotify is now free on mobile, tablet and computer.<br>
+  Listen to the right music, wherever you are.</p>
+</div>
+
+<div class="what-is">
+  <h2>What’s on Spotify?</h2>
+
+  <div class="orga">
+    <div class="col">
+    <img class="grow" src="images/music-icon.png">
```
ta-web-paris: you don't use the class grow in your css. It's better to remove classes that you don't use because  it makes it easier to read the HTML

ta-web-paris:
it looks good!! 

Naming things will come with practice, it's hard! You structured everything well, and I like how you sectioned off your css to make it clear what part you were focusing on 💪

PR #784, opened 2018-10-23
paris-spotify exercise done 

filippofre:


starter-code/css/main.css
```diff
@@ -6,3 +6,140 @@ Green: #00B172
...
+    height: 60px;
+
+}
+
+#spot-app {
+    width: 300px;
+
+}
+
+.p_mr_150 {
```
ta-web-paris: Some of your css names are a little confusing- try to give them meaningful names because it will help a lot when working on a group project. And remember the naming convention for css is kebab-case and not snake_case

ta-web-paris:
good job 👏

it looks good and your code is easy to follow!  I like how you tried to create classes that you can reuse- it will be an important skill during bigger projects

PR #786, opened 2018-10-23
Paris - Régis

jourisque:
Lot of work has to be done but i think i understood what has to be done

### DIFF
--- a/starter-code/index.html
+++ b/starter-code/index.html
@@ -3,40 +3,68 @@
 <head>
   <meta charset="UTF-8">
   <title>Spotify</title>
+  <link rel="stylesheet" href="./css/main.css">
+  
 </head>
 <body>
-Premium Discover Help Download
 
-Music for everyone.
-Spotify is now free on mobile, tablet and computer.
-Listen to the right music, wherever you are.
+  <header>
+    <div class="left-content">
+      <!-- <img src="./images/spotify-logo.png" alt="Spotify logo"> -->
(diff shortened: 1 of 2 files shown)


starter-code/index.html
```diff
@@ -3,40 +3,68 @@
...
 
-Music for everyone.
-Spotify is now free on mobile, tablet and computer.
-Listen to the right music, wherever you are.
+  <header>
+    <div class="left-content">
+      <!-- <img src="./images/spotify-logo.png" alt="Spotify logo"> -->
+    </div>
+    <div class="right-content">
+      <ul>
```
ta-web-paris: you don't need to put them in a list, they can just be in an ``` <a> ``` tag

starter-code/index.html
```diff
@@ -3,40 +3,68 @@
...
+    </div>
 
-Millions of Songs
-There are millions of
-songs on Spotify
+  <div class="green-content">
+    <div class="left-green content">
+      <h2>It's as yeezy as Kanye West.</h2>
+      <h3>Search</h3>
+      <section>Know what you want to listen to ? Just search hit and play.</section>
```
ta-web-paris: they should be in ```<p> ```tags rather than in  a section  tag

ta-web-paris:
it's a good start 💪

If you want to try and finish it, let us know and we'll give you some more feedback on it. css is hard, and it takes practice to master it

PR #787, opened 2018-10-24
AMS jesco wuester

jescowuester:


IronhackAMS:
Looks very good. Try, however, to avoid using id's where you can also use class names.

PR #794, opened 2018-10-24
[MAD DE] Fernando Pérez de Ayala

ferpar:
Exercise complete. The hover event on the button is not working reliably. In some cases, it seems to keep triggering the event constantly until you stop hovering. Why would this be? 

Besides this, I used viewport relative measures everytime I recalled to do so... the kayne west image will not adapt though, cause I used percentages on that one.

Cheers!


ta-web-mad:
Muy buen ejercicio. Las animaciones tienen muchas propiedades. Te paso una posible solución. Fijate en la iteración: inifinita. Para que no pare la animación.
```
.spotify-logo:hover {
  animation-name: spotify;
  animation-duration: 0.2s;
  animation-direction: alternate;
  animation-iteration-count: infinite;
}
@keyframes spotify {
  100% {
    margin-bottom: 20px;
  }
}
```

Sigue asi. 

#checked #Teo :muscle:

PR #829, opened 2018-10-30
AMS David Bros

TheHunter896:
Finished


IronhackAMS:
Nice work. See if you can use class names more generically so you don't have to repeat your code.

PR #941, opened 2019-01-17
DE-MEX-Web_Spotify-Clone_Isabella-Onate

saipasha:
Complete but not finished. Nav bar keeps going below every element that's not directly in the body.


ta-web-mex:
I loved it. Yeah, the last section is over the navbar I suggest trying with z-index :D
🚀 🤖 🚀 🤖 🚀 🤖

PR #950, opened 2019-01-24
Fischer - Jack - FTWD - Miami

fischerjack:
Finished Spotify homepage clone - no frameworks used - not mobile-friendly

### DIFF
--- a/starter-code/css/main.css
+++ b/starter-code/css/main.css
@@ -1,44 +1,181 @@
-/*
-Colors:
-
-Text: 1A1A1A
-Green: #00B172
-White: #FFF
-
-*/
 
 body{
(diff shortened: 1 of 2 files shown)


ta-web-mia:
Everything looks beautiful 🤪
The only thing is that the text in the main image doesn't look centered. You can do this instead:
```
#heroImage > div{
  height: auto;
  /* padding-bottom: 325px; */
  padding: 300px; 
  background-size: cover;
  background-position: bottom;
  text-align: center;
  color: white;
}

#heroImage h1{
  /* padding-top: 200px; */
  font-size: 5rem;
  margin: 0;
}
```
and it will be perfect!

PR #967, opened 2019-01-31
worked with guido

amperry:


starter-code/css/main.css
```diff
@@ -1,8 +1,139 @@
...
+    color: #FFF;
+    padding-left: 70px;
+}
+
+#green-stuff img {
+    width: 300px;
+    position: absolute;
+    top: 130px;
+    left: 800px;
+}
```
lienekechee: Not sure `position: absolute` is ideal here, since the image will stay in the same place in relation to your `#green-stuff` section.  Since you have positioned the `img` from the left it has funny consequences when the width of your screen is smaller.  Perhaps try using the `vw` (view width) unit for a most responsive layout.  And I would say if the image is closer to the right, set position from the right so that if anything gets cut off its the empty bit in the middle

starter-code/css/main.css
```diff
@@ -1,8 +1,139 @@
...
+
+#whats-on p {
+    font-size: 23px;
+    font-weight: 100;
+}
+
+#green-stuff {
+    background-color: #00B172;
+    height: 800px;
+    width: 90%;
```
lienekechee: Be wary of mixing relative values (percents, vw, vh) with fixed values (pixels) when dealing with widths because the results will always be different depending on the size of your screen.

PR #972, opened 2019-02-01
AMS - Isabela

isabeladowsley:


starter-code/css/main.css
```diff
@@ -1,8 +1,185 @@
...
+    left: 25%;
+    top: 30%;
+}
+
+
+.mainbox h1 {
+    font-size: 70px;
+    display: flex;
+    justify-content: center;
+}
```
lienekechee: I think `text-align` is better here. You should add the `display:flex` property to containers with items in it that you want to organize with more of a row/column logic.  If its just text you can use text-align :)

PR #974, opened 2019-02-01
Henry Doce FTWD Miami

hendo12:
Spotify Clone with wonky images


starter-code/css/main.css
```diff
@@ -6,3 +6,107 @@ Green: #00B172
...
+  text-align: center;
+}
+
+.columns {
+  width: 33%;
+  float: left;
+}
+
+.define {
+  height: 500px;
```
ta-web-mia: Don't hard code the height so the content doesn't leak to the div bellow

ta-web-mia:
Good Job!! 😆Fixed what I commented to make it better

PR #982, opened 2019-02-01
Liran Gibor FTWD-Miami 2019

LiranGib:


ta-web-mia:
Good Job!!! To get rid of the white space around the page make the padding and margin of the body equal to zero. Other than that, everything looks good!!!

PR #991, opened 2019-02-04
done

camilav2:


starter-code/css/main.css
```diff
@@ -6,3 +6,150 @@ Green: #00B172
...
+    border-color: white;
+    overflow: visible; 
+}
+
+header .spotify-green-logo {
+    float: left;
+    height: 15%;
+    width: 15%;
+    margin-right: 65em;
+}
```
lienekechee: Because the spotify logo is the only `img` in the header, you can target this using `header img` or make `spotify-green-logo` an `id`

starter-code/index.html
```diff
@@ -1,42 +1,90 @@
...
+    were listening live<br>
+    </div>
+
+    <div class = "stream">
+    <img src="images/devices-icon.png" class="devices-icon">
+    <h3>Stream Everywhere</h3>
+    Stream music on your<br>
+    Smartphone, tablet or<br>
+    computer<br>
+    </div>
```
lienekechee: Keep in mind that classes are meant to be reusable and ids should be unique.  Here you have three boxes that should be the same except for their content.  In this case I would try something like this : `<div class = "box">`
so that in your CSS you can give all your boxes a width of 33.33333%  without having to repeat yourself.  Then if you need to add specific attributes you can target the children of `.column` using the [nthChild selector]( https://www.w3schools.com/cssref/sel_nth-child.asp).  Or you can make the `.column` container a flex container and let it fill its space equally. 

Basically if you can think of better "general rules" to apply to your layout you will have an easier time and have to use few selectors and classes.

starter-code/css/main.css
```diff
@@ -6,3 +6,150 @@ Green: #00B172
...
+    margin-right: 65px;
+    width: 27%;
+    height: 80%;
+}
+
+#green-section>.spotify-white-logo {
+    position: absolute;
+    margin-top: 20%;
+    height: 20%;
+    width:12%;
```
lienekechee: I'm seeing some issues with aspect ratio for your images.  If you remove the `width` properties here and for the `spotify-app` photo, the the size will be adjusted keeping the aspect ratio intact.

PR #1015, opened 2019-02-10
[AMS] Merle Jansen

jurlem:
hand-in version of Spotify clone

### DIFF
--- a/starter-code/css/main.css
+++ b/starter-code/css/main.css
@@ -1,8 +1,140 @@
-/*
+ /*
 Colors:
-
-Text: 1A1A1A
+Text: 
 Green: #00B172
 White: #FFF
-
 */
+
+ body {
+   margin: 0;
+   padding: 0;
+   font-family: 'Open Sans', sans-serif;
+   color: 1A1A1A;
+ }
+
+ nav {
+   display: flex;
+   list-style: none;
+   margin: 20px;
+ }
+
+ .nav-left {
+   justify-items: center;
+ }
+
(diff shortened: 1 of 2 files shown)


starter-code/index.html
```diff
@@ -1,42 +1,71 @@
...
+          <h3>Search</h3>
+          <p>Know what you want to listen to? Just search and hit play.</p>
+          <h3>Browse</h3>
+          <p>Check out the latest charts, brand new releases and great playlists for right now.</p>
+          <h3>Discover</h3>
+          <p>Enjoy new music every Monday with your own personal playlist. Or sit back and enjoy Radio.</p>
+        </div>
+      </div>
+      <div class="small-logo"><img src="images/spotify-icon-white.png" alt="white logo" width="100px"></div>
+      <div class="player"><img src="images/spotify-app.jpg" alt="spotify app" width="80%"></div>
```
lienekechee: Setting a relative width to your items can be problematic, because different screen sizes will have a greater impact on your styling. Since you are using flexbox, try to take advantage of the features that allow for items to be distributed dynamically inside their flex-containers.

starter-code/css/main.css
```diff
@@ -1,8 +1,140 @@
...
+   width: 80%;
+ }
+
+ .text-under h3 {
+   font-size: 29px;
+   padding: 15px 0;
+ }
+
+ .small-logo {
+
```
lienekechee: Watch out for empty spaces :)

PR #1041, opened 2019-03-24
Paris, Miguel Bento

MiguelIronHack:


starter-code/index.html
```diff
@@ -1,42 +1,97 @@
...
+  <body>
+    <nav class="navbar">
+      <div class="nav-cont">
+        <img
+          class="nav-img"
+          src="./images/spotify-logo.png"
+          alt="spotify logo"
+        />
+        <ul class="navbar-content">
+          <li>Premium</li>
```
abisummers: These should be in a tags because they would link you to other parts of the site

ta-web-paris:
You're really good at css, this looks great!! 🙌

PR #1067, opened 2019-04-02
Miami- FSWD-Apr- Justin Velez

jvelez523:
Completed with Jonathan


ta-web-mia:
Good job! In retrospect, you guys can see how the use of flex properties of bootstrap classes could make the icons section more responsive. Nothing you should go back to, but it's definitely something to keep in mind moving forward. Keep it up!

PR #1068, opened 2019-04-02
MIA FTWD April Kenneth Solano

klsolano7:
done


ta-web-mia:
Good job! In retrospect, you can see how the use of flex properties of bootstrap classes could make the icons section more responsive. Nothing you should go back to, but it's definitely something to keep in mind moving forward. 

And just as a quick reminder, you can add the following code to the top of your CSS to get rid of the thin white margin that surrounds the page:

body {
   margin: 0 auto;
}

The reason this works is because the body has a small amount of margin on it by default, so you need to explicitly give instructions to remove it. Overall, very good job with this assignment. Keep it up!

PR #1069, opened 2019-04-02
MIA-FTWD Angel Biondi

AngelBiondi:
Done


ta-web-mia:
Not bad! The main issue I have is that the top section contains an image at the top that's pushing everything down, when it should in fact be a background image. To do so, you would declare a "background-image" property in the CSS for the top section of the page. I would only go back to it if you find yourself with extra time, otherwise just keep it in mind moving forward.

PR #1070, opened 2019-04-02
MIA FTWD - Christian Khawam

CodeByCK:


ta-web-mia:
Good job! The only big issue I had wasn't with the appearance of the page itself, but that the pathing for the background image and the CSS were off. Just a reminder, when you need to path backwards a directory, it needs to be notated with a dot. So from the index.html, you would need to put

`./css/main.css` 

since you can't move forward until you've backed out of the index.html. Either way, the actual page itself looks great. Keep up the good work!

PR #1072, opened 2019-04-02
MIA - FTWD - APR - Nathaniel Carvajal

nathancarv:


ta-web-mia:
You're on the right track! Setting the top section to have a background image was the right idea, and the navbar is definitely correct outside of font properties. I wouldn't go back to this assignment since the stuff moving forward is more important, but feel free to reach out if you need a little extra help with styling. You're always going to be styling to an extent, so definitely do whatever you can to get comfortable with it.

PR #1073, opened 2019-04-03
MIAFTWD - April - Chris C and Nate

ccardelle:


ta-web-mia:
Nice job! Aside from fixing some font sizes in the CSS, there's not too much to go back to. I appreciate that you implemented some flex properties as well. Another way you could have used it was to apply `display: flex` to the div containing the 3 icons and images in the middle section. This would have caused it to wrap as the width of the page got thinner. Keep up the good work!

PR #1074, opened 2019-04-03
MIA-FTWD-Yulia

salapjaka:
done


ta-web-mia:
Good job! The only thing that really can be fixed is the bottom section, which wouldn't be much more than changing some width and positioning. To do so, I would wrap the bottom text in a separate div and reduce the width of it. Then, I could use positioning to move the image of Kanye to the right side of the section. I wouldn't go back to it since there are other more important assignments, but I'd keep it in mind moving forward. Keep it up!

PR #1075, opened 2019-04-03
MIA FTWD ANGHY POMAR

apomarn:


ta-web-mia:
Great job! In retrospect, you can see how the use of flex properties or bootstrap classes could make the icons section more responsive. Nothing you should go back to, but it's definitely something to keep in mind moving forward. Keep it up!

PR #1096, opened 2019-05-29
PAR - FTWD - Virginie

Virbee:
the ratio of the elements in the green section are not pixel-perfect


ta-web-paris:
Good work overall : )

PR #1100, opened 2019-06-01
PAR - FTWD - Cecile Pham

sheshiru:


ta-web-paris:
If you get rid of the position relative on the last image and use the flexbox properties, you will be able to adjust it to the container : ).
Other than that well done.

PR #1101, opened 2019-06-02
Paris - FTWD - Léa

leaandrea:
Could be better with flexbox and positions, will practice.


ta-web-paris:
Its really good though ! well done.

PR #1102, opened 2019-06-02
Iron Hack Paris - Lab Spotify Clone - Yacine Berrada

yacinebe:
Still a few things I would like to fix

Make the navbar fixed while using flexbox
make it reponsive


ta-web-paris:
Nice job.
Try 

.main_navigator {
  display: flex;
  position: fixed;
  padding: 10px 10px;
  width: 100%;
  background-color: white;
}
There will most likely be something to fix in the overall container

PR #1103, opened 2019-06-03
PAR FTWD Souhair 

souhairb:
Je n'ai pas su aligner les éléments du <nav> dans le bon ordre alphabétique
idem pour les éléments dans le footer.

### DIFF
--- a/starter-code/css/main.css
+++ b/starter-code/css/main.css
@@ -1,8 +1,165 @@
-/*
-Colors:
-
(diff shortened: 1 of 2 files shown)


ta-web-paris:
Hey, you should take a look at the flex-box, it'll make your life easier and your page a little more responsive, you can try out the flexbox froggy game to get used to it, come to us so we can get you started with flexbox

PR #1104, opened 2019-06-03
[PAR - FTWD - Octave]

OctaveStoufflet:
spotify done


ta-web-paris:
Closing the pull request, i see you used flex but some other times you're using position float/ absolute, you can handle your page just with flexbox, it will make it easier for you, come to us for a quick brief on structuring your page and using flexbox to make your elements fit in.
Other than that you were pretty close

PR #1116, opened 2019-06-06
LIS - Andre DEV#144 - DONE

anwittw:


carlarsmendes:
Hi Andrew! I loved your css :) You did a great job, even making it partially responsive, so, yesss, flying colors!
I have a couple of notes for you, only for improvement:
1- Naming the classes in CSS is often tricky and you did a good job.
2- The section class="image" should actually be your <header> and your <header> should be your <nav>. And your last section should be your footer . Why should you care about semantic html? It's better for SEO, for accessibility (machines reading your code), and also for another people reading your code.
3- Great job when organizing your CSS. The only thing I could advise you would be to try and find more common classes to write less code.
4- On the  "three-boxes" section, I think that it would've been a better approach i thre were individual divs for each and then position with flexbox. Making them all the same size, also, by defining the width as the same for the three.

Alright, all and all, great job  :100: 

I LOVED THE FOOTER!!! :heart:

PR #1117, opened 2019-06-06
Lis - Seb

smorel14:


carlarsmendes:
Hi Sebastian! 
Just checked your HTML and CSS!
Here are some advises for the future

1- What happenened with your initial HTML <head> ? We advise always to use the ! + tab on VS Code to start your projects so that you keep the meta tags which are important so that your site works properly.
 2-  Also, you <link> to the stylesheet was in a limbo -> between head and <bpdy>. you should not have anything there, always have <link> inside your head (even i it works)
3- I would also advise you to check your classes names'  
image
here, if you already use "nav" then you don't need to add another class to reference it.And  here:
image
This tag should be "header" -> THis is according to semantic HTML. It is best practices and it's better for: SEO, accessibility, and in case others are checking your code.
4- Why do you have "alt"= "error" ? The point of "alt" is for screen readers for accessibility purposes. Because some people would not be able to see the image, this would be the place to leave a small description of the image. 
5- always check your VSCode for errors, like these:
image
And make sure to correct them before doing your final submit
6- On your navbar, always use ul and li even if you're going to use flexbox - > Since it is a list of items
7- I noticed you had a small margin on your website. for next time, this will help you:
image
when you do this, you reset your browser's inherent properties and can start using your own.
8- Always try to separate your CSS in sections, like this:
image

And of course, you missed the last section which was also tricky ;) 
But I'm sure you'll practice with the slack lab  :P

Good job!!! :+1: :100: :+1:

PR #1118, opened 2019-06-06
LIS - Sonia #144

soniameller:
I had a lot of trouble with positioning


carlarsmendes:
Hi Sonia, You did a great job :) (despite your comment on positioning)
I'll just leave you some feedback for improvement:

1- Always check your visual code for warnings
image
2- Use semantic tags: <nav> for your navigation bar, <header> for the first sections when you have h1 and h2, and sections instead of divs :) Why should you sue them? For SEO purposes, accessibility, and for readibility -> Basically, best practices. 
3- On section-2 , I would advise you to give classes to your divs so that you could style them altogether and then use flexbox.
image
Also, it avoids to have this : #section-2>div>div>img  -> which can get very confusing once you have bigger projects. You should also always think that you can have other people checking your code. You should be more specific when naming your classes, according to the content inside -> Those are best practices
4- For background, practice to using the shorthand property, like you can see below:
   "body { 
  background: lightblue url("img_tree.gif") no-repeat fixed center; 
}"

Thats it, youd did a great job, continue with positioning.
Here two cool links:
http://flexboxfroggy.com/
https://flukeout.github.io/

:+1: :100:

PR #1119, opened 2019-06-06
[LIS] Giulia - done with doubts

GiuliaGallorini:


GTCarmona:
Heeeey Giulia!

Fist of all, really proud to see your commitment to finish this exercise. I know it's hard and it takes a lot, but this is how you learn and improve!!

your exercise is really good and close to the original! I like how you used the red border to visualize each div and you are naming the classes in a very specific way, that is good and makes it easier to read the code
Here are some notes to make it even better
- In your "container-features" class, you can increase the width of the div so the features have more space between them.
- About the List items on the navbar, that we were discussing before. Don't feel like you are "cheating" by adding some margin or padding. Almost every tag will have some padding or margin so it's really important to understand how and why you should use them, and they are very important to make things better positioned and visually appealing. 
In this case, I'll show you how I would write them to make them cleaner and work just as well

This is what you did:
---------------------------------
.spotify-logo {
    height: 100%;
    width: auto;
}
.nav-list {
    display: flex;
    flex-direction: row;
    justify-content: space-around;
    margin: 0 10px 0 0;
}
.nav-item {
    text-decoration: none;
    list-style: none;
    margin: 0 5px;
}
------------------------
This is what I would do:
.spotify-logo {
    height: 100%;
}
.nav-list {
    display: flex;
    list-style: none;
}
.nav-list li {
    padding: 0 3vh;
}
------------------------
You dont need a class for every single tag if you are able to access them through it's "parent" tag, like I did here with the "li" tag. This makes your CSS less confusing and better to read.
That is just a few tips, keep working hard and soon you'll see the results! if you need anything let us know!

Guilherme

PR #1132, opened 2019-06-07
LIS - Carlos - almost done

carlosbotto:


GTCarmona:
Hey Carlos,

I see that you are still working on it. Let me know in class or on Slack when you are done with it (either finished or you decided not to touch it again)
Don't focus on the responsiveness, try to put things in their correct place using Flexbox. And if you have any questions feel free to call me or Carla :)

PR #1136, opened 2019-06-10
LIS - Anja - finished Spotify clone

anjaboettcher:


carlarsmendes:
HI Anja! first of all, good job, looks good

- Good job with your classes, I think you did a good job! It's not always easy to name them but you did you did it correctly -> more according to your content/position, than to colors, for example
- good job with picking up a google font :)
- And with flex box!!
- And with the background image

I'll just give you a few tips for next time:

1-  Always check your errors on VS Code - here for example it was saying that you had an unclosed div
image
(I think you forgot your " class="right-side ">")

2- Use the semantic HTML tags ;) in your case, <nav></nav> for the nav bar, <section> for the various sections;

3- You can always use the shortcut ! + 'tab' to start your HTML document. The advantages of this is that you already have some predefined meta tags you don't have to think about anymore, and you don't get confused by them;

4- When you do your nav bar as a list ( <ul><li>) you don't have to use any additional classes and that will probably help you, so check it out! still, what you did was correct, it's just more usual to build the nav bar as an unordered list :) (by naming your items 'view', I don't think another person reading your code would understand that those were your nav items)

5- Your "part-1", I would tag as "header" -> Since it contains the h1 and h2 titles

6- I saw you used 'hr' for the underline, you could also use 'border-bottom' and style it accordingly 

7- Use shorthand properties, for example, for background and margin. It will shorten your code

8- Divide your CSS into sections, this will help you organize yourself in the future
image

That's it! Really good job Anja, let me know if you need anything else :) :+1: :100:

anjaboettcher:
Thanks so much for your feedback, Carla! I will implement your corrections! :)

PR #1162, opened 2019-06-12
MIA-FTWD-Artem

arstrel:
Job is done


MarcosPalaciosIII:
Great job on the spotify clone. CSS is and you will always learn something new if you dive deep enough into it. Using positioning is something that is very helpful when styling as there are times when you want things to be in a certain position and basic CSS just wont give you what your looking for.

Keep up the great work.

PR #1163, opened 2019-06-12
MIA-FTWD-labs-css-spotify-clone-Pablo

PabloVazquezFayat:


ta-web-mia:
Looks good! The navbar doesn't scroll with the page, but everything is in the correct position, nice job!

PR #1165, opened 2019-06-13
MIA-FTWD- Andre Cabrera

andregcab:


MarcosPalaciosIII:
Good attemp on the spotify clone. CSS is and you will always learn something new if you dive deep enough into it. Using positioning is something that is very helpful when styling as there are times when you want things to be in a certain position and basic CSS just wont give you what your looking for.

Try not to mix up displays with positioning, and always try to set relative paths when linking images or files n folders. You CSS and images had broken links or weren't connected. Remember ./ is to stay in the folder your in and ../ is to go out one tier in your folders.

Keep up the good work.

PR #1166, opened 2019-06-13
MIA-FTWD-SERENA DOMINGUEZ

SDominguez21:


MarcosPalaciosIII:
Good attemp on the spotify clone. CSS is and you will always learn something new if you dive deep enough into it. Using positioning is something that is very helpful when styling as there are times when you want things to be in a certain position and basic CSS just wont give you what your looking for.

.Aalways try to set relative paths when linking images or files n folders. Your  images had broken links or weren't connected. Remember ./ is to stay in the folder your in and ../ is to go out one tier in your folders.

Keep up the good work.

PR #1170, opened 2019-06-13
MIA FTWD Matiny Louis

Matiny:


ta-web-mia:
Well done! Not much to critique on my part, it's even responsive. I will say that bootstrap classes/media queries would likely improve the readability on mobile than just having it shrink, but with the amount of time you were given I don't think you can do much more than this. Keep up the hustle 👍 
- Nick

PR #1171, opened 2019-06-13
MIA-FTWD-Nicolle

nicollegonzalez:


ta-web-mia:
Good job! I think with the new Flex properties we've went over, you can improve the icon cards in the middle section. No need to go back to it, but definitely something to keep in mind. Keep it up 👍 
- Nick

PR #1172, opened 2019-06-13
mia-ftwd-jun2019 - Billy Wong

billywonka3:


ta-web-mia:
You have everything in the right place, but the images do not fit the full page. The navbar also doesn't move while scrolling down the page. But this looks pretty good, I'm glad you got the three images in place in the second section

PR #1176, opened 2019-06-13
MIA FTWD - Taluama Velludo

taluama:
Please come talk to me about the font weight when you have a chance. It was working and then it stopped out of the blue and I couldn't figure out why.

Thanks!


ta-web-mia:
Don't worry too much about this, but I will keep this open. I'm not sure what the issue is, this looks very good, just increase the height of the kanye-section to catch the cut-off text.

taluama:
OK, thanks! Just fixed the height.

Talu

ta-web-mia:
Nice job, I'll close this now!

PR #1179, opened 2019-06-14
MIA-FTWD Anabel Perez

anabelpdv:


ta-web-mia:
Nice work! If you have time, try and reposition some of the elements on the bottom section (the images on the right) to get everything a little closer to the original you're cloning. Keep up the good work 👍 
- Nick

PR #1180, opened 2019-06-14
MIA PTWD Antquanique Dancy

adancy95:
Hi, I am having trouble with the nav bar. When I floated the nav links to the right, the order reversed. I attempted to the flip the order by moving the float: right to the actual nav element. However, it did not do anything. So, my work around was to reverse the order of the nav links in the html file.

### DIFF
--- a/starter-code/css/main.css
+++ b/starter-code/css/main.css
@@ -6,3 +6,172 @@ Green: #00B172
 White: #FFF
 
 */
+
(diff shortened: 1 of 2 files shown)


ta-web-mia:
Nice job! Now that you've learned flex and Bootstrap, you should be able to make pages like this even more responsive. Keep up the good work 👍

PR #1181, opened 2019-06-14
MIA_FTWD- Sams Diana 

sdiana11:


ta-web-mia:
This is good, but I'd give a couple notes: set the margin and padding to 0 on the whole page at the beginning of CSS. There is also a grey bar at the top for some reason. Also the navbar doesn't scroll with the page. These are all common things, so don't worry too much, just try and fix these for the next lab!

PR #1182, opened 2019-06-14
MIA-FTWD-Candido Concepcion  -Missing bottom part

candidocon:


ta-web-mia:
So far so good! What you have so far looks really good, and I want you to stay focused on more recent labs. I would suggest you only go back to this if you have extra time after finishing the other ones. Of course, it's up to you! Keep up the good work 👍 
- Nick

PR #1183, opened 2019-06-15
MIA-PTWD-Alexander-Hoinville

ahoin001:


ta-web-mia:
You're definitely on the right track! With just margins, positioning and floats, it can be a little tricky to position elements exactly the way you want them. Now that you've learned about Flex properties though, a lot of this should be a little easier. Keep up the good work 👍

PR #1184, opened 2019-06-15
MIA PTWD Alejandro Salas

alejotsv:
A was not able to fix the top nav without overlay when scrolling down


ta-web-mia:
Pretty good! I know this late, but you can fix the navbar by just adding about 100px of margin-right to the unordered list. Other than that though, everything looks good. Keep it up 👍

PR #1186, opened 2019-06-16
PTWD MIA - Jesus Garcia

JesusGarcia23:


ta-web-mia:
Definitely on the right track! You've learned a lot more about styling since this lab, and you should be able to use a lot of it to your advantage now. Do what you can to learn from this, and keep up the good work 👍

PR #1188, opened 2019-06-17
finished

GulfiyaIsyangulova:


ta-web-mia:
Nice job! Now that you're more familiar with flex properties and bootstrap classes, you can make sections like the middle and bottom ones more responsive. Keep up the good work!
- Nick

PR #1189, opened 2019-06-18
PTWD - MIA - Nelly Gonzalez-Vinas

nellygvinas:
Had some reservations about sizing the image and page content to fit my screen. Once the navbar was set to fixed, the Navbar is wider than the images and content.

### DIFF
--- a/starter-code/css/main.css
+++ b/starter-code/css/main.css
@@ -1,8 +1,252 @@
 /*
 Colors:
-
 Text: 1A1A1A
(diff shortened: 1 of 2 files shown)


ta-web-mia:
Nice work! The extra width added to the page comes from your icons section under the landing page; you added some extra padding-left and padding-right (250px each) that extended it off the side of the window. Other than that, everything seemed to be spot on. Keep up the good work 👍

PR #1190, opened 2019-06-18
MIA-FTWD Steven Englander

confidenti-al:
Completed. Tried a few different variations of the underlines, also tried to use the CSS approaches from pre-flexbox lessons.


ta-web-mia:
A couple of the section titles with 2 lines are overlapping each other vertically, but otherwise, it looks good. Just try to fix those places.

PR #1191, opened 2019-06-18
MIA-FTWD ALFREDO ROJAS

Alfredo-Rojas:


ta-web-mia:
Pretty good! Your top sections look really good, and you were definitely on pace to wrap up the last parts as well. I think my only suggestion is to make sure you feel comfortable with relative positioning, even with your newfound Flexbox and Bootstrap knowledge. Stay sharp, and keep up the good work 👍 
- Nick

PR #1192, opened 2019-06-19
Jose De la Cruz MIA-PTWD

jdelacruzjose:


ta-web-mia:
Your page looks really good when it's viewed from 25% zoom, but everything is extremely large when it's on normal zoom. Normal zoom for reference:

image

And at 25%:

image

Just make sure that your local version of an app is at the proper zoom when you start designing, otherwise it might look very weird on other devices. Keep up the good work 👍

PR #1202, opened 2019-06-21
MIA PTWD Randolph Amaya

Randolphjr7:
Finished Assignment


ta-web-mia:
Pretty good! I think some of your sections could use a little bit more padding to have them more closely resemble the original, but everything else looks great. Keep it up 👍

PR #1237, opened 2019-08-06
PAR FTWD Sara-Loup

SebHacker:


ta-web-paris:
Good, you could improve the nav just a little bit so that the logo is aline to the text (align-items :center) and the text is not stuck to the border of the window(add padding on the sides?)! but it's a detail :)

PR #1239, opened 2019-08-07
PAR FTWD Xavier & Justine

Juddorange:


ta-web-paris:
Nice work. There's juste a complicated thing happening with the 3 col icon section. The grid seems a bit weird. You can come to me if you want to have a look a it more precisely

PR #1240, opened 2019-08-07
[PAR - FTWD - Julie]

juliecarra:


ta-web-paris:
nice one.
Don't forget :
body{
margin:0
}
It's the reason why you have a default margin on the left of your page : )

PR #1241, opened 2019-08-10
PAR FTWD – Antonin – CSS Spotify Clone

gramsco:


ta-web-paris:
Nice work, it is good that you tried to to the responsiveness but it is not really what a responsive layout would react (come and ask me if u want => mimi)

PR #1243, opened 2019-08-14
LIS- FTWD-PaulaBrinkhege-#156

PaulaMarcella:
I had trouble organising the css in a clean way (i feel i have used too many classes and ids)
i need feedback on how to make the website responsive for mobile


soniameller:
Hey Paula!
### CSS classes
Regarding your doubts about css classes, you can read [this article](https://css-tricks.com/the-difference-between-id-and-class/) that is very clear and includes most of the things I can see you are confused about regarding CSS. 

An important point in it  is that **If you don't need them, don't use them**.
#### Nav bar
 - You `log-div` class is not making any differences. I just changes a bit the spacing between the logo but I think it is an undesired effect. If you erase that whole class the page is working perfectly!
- When naming the class in the links `anchor-link` you are repeating something html is already giving you. The `<a></a>` tag is already defining an anchor-link so you could use the tag to modify style. If you need more specific styling you should be more specific in naming classes too. In this case you are not even using the class in the stylesheet, so again, *if you don't need them, don't use them*
#### Green section
 - It would be easier to position the elements if the white logo was a background, then it wouldn't be taking any space.

### Responsiveness
It is always useful to understand what's going on in your project to draw an outline that wraps every item by adding it to your universal selector
```
 *{
    margin: 0;
    font-family: 'Cabin', sans-serif;
    line-height: 2em;
    outline: 1px dashed red
}
```
Your navbar is working with a nice responsiveness, I see you are having trouble with your `info-box` and your `green-box`.
Flexbox is a bit tricky but once you understand it in makes it pretty easy to archive the behaviour you need. What it is important to get is that flexbox has two ways of managing things. Either you have a container or you have items. You can also have container inside containers but you should understand when an element is a container and when an item, for example: you shouldn't display flex` id="log-div"`, this is an item inside the `nav-bar` container.
The [documentation](https://css-tricks.com/snippets/css/a-guide-to-flexbox/) has two columns (left: container, right: items). I recommend you to always have the tab opened when you are using flexbox!

The following is a print screen of some changes I did in your info-box
Captura de Pantalla 2019-08-19 a la(s) 11 55 27

You can check in the [documentation](https://css-tricks.com/snippets/css/a-guide-to-flexbox/) that when you want elements to drop down when they don't fit you can use flex-wrap (in the container) and a possible solution is to fix the width of the items so that they always look the same! 
You could use something similar to solve the problem in the green section.

I hope this was helpful. Let us know if you have any doubts 🙂

PR #1248, opened 2019-08-15
[BER-FTWD-Kateryna_Stadnyk_Iouri_Sorokine]

KaterynaStadnyk:


brudolce:
Flexbox use is fine, but it is highly recommended to use responsive measure units instead of pixels. There was a tidy use of em, keep it.

PR #1249, opened 2019-08-15
[BER-FTWD-Moritz and Marko]

MoCode:
Please, i would be happy about a feedback :D

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+    "liveServer.settings.port": 5501
+}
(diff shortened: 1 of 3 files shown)


brudolce:
Very interesting margin classes. Good measure unit pick for those: em.
There are a few other measure units that are responsive, it is highly suggested to use them instead of pixels. For height you can use vh (e.g. height: 50vh) and for width vw (e.g. height: 50vw); they take in consideration the size of your screen, 100vh should be equals to your whole screen height.

brudolce:
good use of flexbox

brudolce:

PR #1251, opened 2019-08-15
[BER-FTWD-NINETTE-ALFONSO]

fonsogms:


Svemakawe:
Glad that you started to get rid of float and started to use flex in the end:) ! 
I would put the little white icon in the green box as a background-image (easier to center).
Some spacing between your links by putting  **margin-left: 0 7px**  in  **#navOpts a** .

PR #1252, opened 2019-08-15
BER-FTWD-Jörg

Jrgsto:


Svemakawe:
Looks fine (good approach on getting space between the h2 and the undeline-style)! 
Using a google-font could've been nice and to center the little icon in the green box, it's maybe easier to have that as background-image (of the green box).

PR #1253, opened 2019-08-15
Berlin-FT-15-08-2019-Nuno G.

nunogervasio:


angminsheng:
Checked. Well done! Clean code and good usage of flex box.

PR #1254, opened 2019-08-15
LIS-FTWD-[TiagoSilva] #156

TigerSilver:
It is still unfinished but I've been able to go through it, step by step.
Will continue during the next days.


soniameller:
Sure!

You could use outline in the general selector to make it easier to visualize what you are doing.
```
* {
  box-sizing: border-box;
  outline: 1px solid red
}

```

This will create a border in every element and will help you understand the space they are taking. When you don't need them anymore you can comment it out.

PR #1258, opened 2019-08-15
BER-FTWD-Nelly

nellycschneider:


angminsheng:
checked. Good job! Looks exactly the same as the task. Good usage of flex box as well.

PR #1259, opened 2019-08-15
almost finishing spotify clone-daniel Santos

DanielJorgeTeloFilipeDosSantos:


soniameller:
Dani, the same I commented in the clue lab about the naming convention. Please pay attention because it is important to have files clearly organized. 

###  LIS-FTWD-Daniel Santos-#156

Then in the comments you could add a descprition like the one you did in the title  "almost finishing"

PR #1260, opened 2019-08-15
LIS-FTWD-Andre Almeida #156

andreH2M:
Since i have so many doubts i will need the weekend to gain some insights and delivery the best result possible.

### DIFF
--- a/starter-code/css/main.css
+++ b/starter-code/css/main.css
@@ -1,8 +1,26 @@
 /*
+* {
 Colors:
-
 Text: 1A1A1A
 Green: #00B172
 White: #FFF
-
+}
 */
+a {
+    font-size: 20px;
+    font-family: Arial, Helvetica, sans-serif
+}
+
+nav {
+   
+  height: 500px;
+  max-width: 1000px; 
+  display: flex;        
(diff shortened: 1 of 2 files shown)


soniameller:
No problem! I will give you some tips so that you start with less doubts.
- In the code you pushed there is a syntax error. You are opening a <heading> tag but you are never closing it. Remember tags must be paired.  A good practice to check this problems is to keep the vs code console opened. When I opened your code I saw the following message:

```
Tag must be paired, missing: [ </header> ], start tag match failed [ <header> ] on line 20.
```
This image is a screenshot from the bottom of vs code. You can see there is a warning. If you click on the warning the problems tab will be opened with the message I showed before
Captura de Pantalla 2019-08-15 a la(s) 19 26 46

soniameller:
Hey Andre!
- There were still some tags that were not closed. That was generating some strange behaviour. 
Take specific care to close every tag you open and always check if you have any warnings in your VS code.  
- Regarding the general structure of the code. I would rather show you some things whenever you are free. It is going to be more clear that giving a written feedback, please reach me!

PR #1261, opened 2019-08-15
BER-FTWD-Ali-Sarah

SarahHouben:


angminsheng:
Checked. Clean code and looks exactly like the picture. Well done!

angminsheng:
One improvement could be the naming of the classes. The idea of using "section-" for different section is good. But if you changed the  class name to be more meaningful instead of numbers like "section-1" and "section-2" to "section-header" "section-footer" etc, then it will be easier to see which part is the section about. It is also a good practise to use "container" class for the outermost div of each sections.

PR #1264, opened 2019-08-15
LIS-FTWD-Elyas-#156

Essassani:
A 90% accurate copy of the spotify website.
Please feedback on the methods that i used and the amount of divs and classes. Could this be done in a shorter way? In my feeling I made an unnecessary excess of divs and classes to bypass some position methods that could be done easier.  But not sure..

### DIFF
--- a/starter-code/css/main.css
+++ b/starter-code/css/main.css
@@ -6,11 +6,13 @@ Green: #00B172
 White: #FFF
 
 */
-* {
-    outline: 5px solid red;
+*{
+    outline: 3px solid red;
 }
+@import url('https://fonts.googleapis.com/css?family=Roboto:100,300,400&display=swap');
 body {
     margin: 0 auto;
+    font-family: 'Roboto', sans-serif;
    
 }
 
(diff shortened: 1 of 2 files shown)


GTCarmona:
Hey Elyas, 

For sure! I can give you some tips, but you got the logic right, the idea of using the tag selectors is great, and then you can avoid unnecessary classes 

instead of having this class 

```
.free {
    font-size: 1.7em;
    font-weight: 100;
    text-align: center;
    margin: 0;
}
```
 you could have done this

```
.centertext p{
    font-size: 1.7em;
    font-weight: 100;
    text-align: center;
    margin: 0;
}
```
And this concept could be used even for images, so instead of having an ID for the images you can say  
`.right-panel img {}`

same thing with underlined tag

you have this

```
#underlined {
    border-bottom: 3px solid #00B172;
    padding-bottom: 5px;
    font-size: 35px;
}
```

and could be this 

```
underlined {
    border-bottom: 3px solid #00B172;
    padding-bottom: 5px;
    font-size: 35px;
}

```

 So what I would say is, if you have a specific tag that you know you can target, and it is being wraped by a div / section / article , you can keep doing like this and will avoid useless classes


those are all the opportunities I noticed that you could do to improve, but still, is looking awesome!!

PR #1292, opened 2019-08-17
LIS-FTWD-[DANIEL VALERO] #156

damvalero:
done, waiting for feedback to get better the user experience

### DIFF
--- a/starter-code/css/main.css
+++ b/starter-code/css/main.css
@@ -6,3 +6,175 @@ Green: #00B172
 White: #FFF
 
 */
+
+@import url('https://fonts.googleapis.com/css?family=Roboto+Condensed:700|Roboto:100,300,700');
+
+*{
+    box-sizing: border-box;
+    margin: 0;
+}
+
+body{
+    font-family: roboto;
+}
+
+h1{
+    font-size: 70px;
+    color: white;
+}
+
+h3{
+    font-size: 30px;
+    color: #00B172;
+}
+
+.paragraphHeader{
+    font-size: 35px;
+    color: white;
+    width: 820px;
+    text-align: center;
+}
+
+nav{
+    display: flex;
+    align-items: center;
+    justify-content: space-evenly;
+    background-color: white;
+    height: 70px;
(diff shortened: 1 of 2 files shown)


soniameller:
Hey Dani! It looks good 🙃
This kind of exercises have a lot of ways of being solved. I am going to give you some of my suggestions. 

- It is nice to have comments in the code but it is even better if the code explains itself. Instead of having a `container1` and `container2`, give them a name that can help you recognize that sections. 
- You should try to have the least amount of personalized classes and reduce as much as posible the size of your stylesheet. Tags usually give you solutions for the main styling, try to choose the tags that give you the correct height, use the "Cascading" benefits of CSS for, as an example, white fonts.
- For the underlined headings you used `text-decoration: underline. It is a good solution  but, being picky I can tell you that it is not similar to the underlining style of the example. Another common way of underlining things is adding a bottom border and this will draw a line under the word instead of going through it. 
- For **Container 2** you set a fixed height but your content is taking more space. This is generating some weird behaviour. This is a print screen from your code
Captura de Pantalla 2019-08-19 a la(s) 10 25 20
If you don't set a height, the container takes its content dimensions like in the following image
Captura de Pantalla 2019-08-19 a la(s) 10 25 40

   I usually use in the general selector the line: `outline: 2px dashed red`, this draws a line around each of your elements and makes it very easy to understand what's going on. 

- A bonus thing could be to make it responsive. Flexbox is a bit confusing in the beginning but once you understand it it makes it very easy to make things responsive. If you have time and would like to try it out you could use it to make your page look nice in mobile too or you could do the instagram clone just to try something different from scratch! Have always the [flexbox documentation](https://css-tricks.com/snippets/css/a-guide-to-flexbox/)  opened in a tab. It is very helpful!

Reach me or Gui if you have any doubt!

damvalero:
Hello Sonia,

Thank you so much for taking the time to provide me a lot of feedback.
Yes, I will use these corrections that you send me to work in the exercise
of Instagram.

Regards.

PR #1295, opened 2019-08-19
MIA-FTWD-Bobby

bobbypwang:


MarcosPalaciosIII:
Great work on the spotify clone. Positioning can be quite tricky at times. Luckily there are many tricks in css that can help you get the positioning that you're looking for. Good thing is that you'll have lots of practice with this and will in time become much easier.

Keep up the great work.

PR #1297, opened 2019-08-19
MIA-FTWD-Alejandro Gomez

alegomez1:
First draft

### DIFF
--- a/starter-code/css/main.css
+++ b/starter-code/css/main.css
@@ -6,3 +6,138 @@ Green: #00B172
 White: #FFF
 
 */
+
+*{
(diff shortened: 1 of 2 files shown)


alegomez1:
Added hover animations for the menu bar

ta-web-mia:
So everything looks good, but the Spotify picture isn't showing. After looking at your code, you copied exact path to the picture, which will not work on any other computer but yours. You need to find the relative path. I'm going to keep this open, and see if you can fix this and git push again.

alegomez1:
Updated the image links

PR #1298, opened 2019-08-19
MIA-FTWD-ColeMooney&JosueAcevedo

colemooney:


ta-web-mia:
Nice work! And now that's you've learned flex properties, there's a lot you can do to make a page like this even more responsive. Keep it up 👍 
- Nick

PR #1301, opened 2019-08-19
In Progress

marven250:


ta-web-mia:
Looks great! Now that you've learned more about flex, properties like that would be very appropriate for the icon section in the middle. Keep up the good work 👍 
- Nick

PR #1302, opened 2019-08-19
MIA-FTWD-Enrique-Narciso

e-narciso:
Spotify Clone non-reactive design


ta-web-mia:
Nice job! The picture of Kanye is a little too far of the green background, but everything else looks great!

PR #1304, opened 2019-08-20
MIA - FTWD - Christian and Gabriel 

ChristianRoque:


ta-web-mia:
So most of this looks good, but the main spotify picture isn't showing. I looked at your code, and you used the complete path to the image, starting with your local source files, rather than a relative path. No one outside of your computer will be able to see the image. Go ahead and fix that, and I'll close this out.

PR #1305, opened 2019-08-20
MIA CSS SPOTIFY - Gabriel (partner: Christian)

gabrielnes:


ta-web-mia:
Nice work! The logo on the bottom section isn't really cooperating, but don't worry too much about that. You could use absolute positioning to get it right in the middle of the other two block elements, but that isn't necessarily best practice. You've learned more responsive design tools, and I'm sure one of them could provide a more elegant solution. Keep it up 👍 
- Nick

PR #1307, opened 2019-08-20
MIA-FTWD Osheit Chang

osheitchang:


ta-web-mia:
You're on the right track! I think once you resize the elements, everything will fall into place. If you have a chance to go back to this, you're welcome to implement the flex properties that you've learned. Keep it up 👍 
- Nick

PR #1317, opened 2019-08-20
MIA FTWD Jorge Luis

jorgeluis42:


MarcosPalaciosIII:
Good work on the spotify clone. Positioning can be quite tricky at times. Luckily there are many tricks in css that can help you get the positioning that you're looking for. Good thing is that you'll have lots of practice with this and will in time become much easier.

It also seems like you have some broken links to images. Remember to always set relative path from the folder the file is in to the file where the image is at in order to link images and files as well. ie: instead of start-code/images/spotifyLogo.png -- you can do images/spotifyLogo.png
Since images is in the same folder, you don't necessarily have to path out and back to it, just target the folder.

Keep up the good work.

PR #1320, opened 2019-08-21
mia ftdw Jennifer Muniz & Elba Chimilio

Jenniferrmuniz:


ta-web-mia:
You're on the right track! If you have a chance to go back to this, definitely consider implementing some of the flexbox properties that we learned about the other day. Keep up the good work 👍 
- Nick

MarcosPalaciosIII:
Good work on the spotify clone. Positioning can be quite tricky at times. Luckily there are many tricks in css that can help you get the positioning that you're looking for. Good thing is that you'll have lots of practice with this and will in time become much easier.

Keep up the good work.

PR #1402, opened 2019-10-16
[PAR-FTWD] Thomas - Spotify

ThomasAlbrt:
This is my spotify lab


ta-web-paris:
Looks good! Just make sure your route is correct for your background image, otherwise, it won't display well each time --> CSS  line 42  background-image: url(../images/landing.jpg); (two dots)

PR #1404, opened 2019-10-16
PAR WDFT ANDY

N10T:
Spotify's done
Do i need to use id or classes for identify section ?

### DIFF
--- a/starter-code/css/main.css
+++ b/starter-code/css/main.css
@@ -1,4 +1,5 @@
 @import url('https://fonts.googleapis.com/css?family=Montserrat&display=swap');
+
 /*
 Colors:
 
@@ -8,6 +9,8 @@ White: #FFF
 
 */
 
+/* ---------------- General ---------------------------*/
+
 * {
     font-family: Montserrat, Helvetica, Arial, sans-serif;
     margin-top:0;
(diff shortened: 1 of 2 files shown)


starter-code/css/main.css
```diff
@@ -6,3 +8,177 @@ Green: #00B172
...
+/* ---------------- Navbar ----------------------------*/
+
+.navbar {
+    z-index: 10;
+    background-color: rgba(255, 255, 255, 0.98);
+    position: fixed;
+    display: flex;
+    align-items: center;
+    height: 100px;
+    width: 100vw
```
ta-web-paris: for the width: it's better to write 100% instead of 100vw (otherwise an horizontal scroll bar appear)

starter-code/index.html
```diff
@@ -1,42 +1,81 @@
...
+      <div>
+        <h4>Search</h4>
+        <p>Know what you want to listen to?<br>
+          Just search and hit play.</p>
+      </div>
+      <div>
+        <h4>Browse</h4>
+        <p>Check out the latest charts,<br>
+          brand new releases and great<br>
+          playlists for right now.</brand>
```
ta-web-paris: I think you wanted to write "br" and not "brand" ??

ta-web-paris:
As a section would be unique it's good to add an id. Nonetheless you could also add an id + a class, if some of you css rules apply to several section. 

Great Job!

PR #1416, opened 2019-10-17
PAR WDFT [Amine]

aminelahrichi:


ta-web-paris:
So, looks perfect! But only if you dezoom the browser tab (33% of original size). You probably had your browser not on 100% when you made the page. Happens, no big deal :)
Capture d’écran 2019-10-18 à 09 39 28

PR #1421, opened 2019-10-17
[PAR-FTWD]Camille

Millcat:
Spotify finished !


ta-web-paris:
Good job - couple things, the concert picture is supposed to take up the whole screen in width (no white margins), also the nav bar is supposed to be position: fixed; but as long as you understand, all good :)

PR #1422, opened 2019-10-18
Final version Spotify - Parina

ParinaKh:
Please find here the final version of my work.
The website in almost responsive... and I could'nt find the same fonts.

Let me know if it's all good ;)


ta-web-paris:
Very nice work! Responsiveness isn't easy :)

PR #1427, opened 2019-10-21
PAR WDFT - Marie

Mawee33:
Done with 2 errors : 
- the underlines in sections 2 and 3
- the ending images sizes


ta-web-paris:
Good job! For the last section, putting classes on your div, as well as height and width in the CSS might help :)

PR #1428, opened 2019-10-21
PAR [WDFP] Marion

marionboccaccio:
Hi guys it is almost done :-) but not finished :-(


ta-web-paris:
Watch your routes, "./images/spotify-logo.png", no need to go back to the starter-code level :) But as long as you practice flexbox, no need to spend more time on this :)

PR #1429, opened 2019-10-22
rendu mais avec pleins de modifications a faire

NoemieNT:


ta-web-paris:
Make sure you practice Flexbox (codepen --> colored divs into divs, play with flex-direction etc) but no need to spend more time on this lab! Focus on JS, though don't hesitate to ask us if you need clarification on the HTML/CSS part :)

PR #1571, opened 2020-01-07
[PAR-WDFT] - TREBOIT ongoing

franck-treboit:
This is my first approach. I feel some things are not ok but close to be. 

Waiting for your advice !


ta-web-paris:
Pretty good job!! Be careful with your relative paths though, when inserting an image that is in a subfolder, for example "src="/images/music-icon.png", you need to put a dot before the "/", like so: src="./images/music-icon.png". The dot means "starting from here" and it will cause the images not to display well everywhere if you forget it :)

PR #1572, opened 2020-01-07
[PAR-WDFT-0120] - Yannick Bourenane - ongoing

yannick-bourenane:


ta-web-paris:
Great job ! The site looks great and very similar to what was asked !
I also really like the fact that you keep your colors on top, very good practice ;)

2 things :  
- `body {
  font-family: Ubuntu, Helvetica, Arial, sans-serif;
  color: #1a1a1a;
}`
The font-family used will always be the first one, here the Ubuntu is not recognised so your full page is writed on Helvetica. You can erase the other ones as they are only taking memory.
- Your navbar is wrapped in a lot of element tags, I would combine your `wrap-flex` and `topbar` divs into a `nav` tag, remove the `ul` and `li` tags and simplify your css.

Anyway keep up the good work !

PR #1573, opened 2020-01-07
[PAR-WDFT-0120] Pierre Etryhard ongoing

PierreEtryhard:


ta-web-paris:
Almost good! You need to work a bit more on the last section, for example there are no margins around it.  Also, there might be some elements making your page too wide (you're not supposed to have an horizontal scroll bar) - come to us (clara) is you have any questions ! :)

PR #1575, opened 2020-01-07
[PAR - WD - FT] JONATHAN OREJA ongoing

Jarchiii:


ta-web-paris:
Pretty good clone, but make sure you modify in the original index.html and main.css files, because you had two of each and it makes the relative paths buggy ! :)

PR #1577, opened 2020-01-07
[PAR-WDFT] - Alix ongoing

AlixdeR:
done


ta-web-paris:
Almost perfect ! You would just need to add "z-index: 1" into the CSS properties of your nav bar to make it go over the last green section when you scroll (it shouldn't disappear under it).

PR #1579, opened 2020-01-07
[PAR-WEBFT] ALEX MEYER ongoing

ameyerfr:


ta-web-paris:
Good job! The only thing you'd need to make the first section bigger (need to take up the whole screen). But it's a detail so I'll close the pull request.

PR #1580, opened 2020-01-07
[PAR-WDFT]-PASCAL-BESSON - ongoing

pascalbes:


ta-web-paris:
Good job !

There is still some things to improve :

in this block of code : 
> `@import url(https://fonts.google.com/specimen/Montserrat?selection.family=Montserrat)`
> 
> `*` {
>     margin:0;
>     padding:0;
> }
> 
> body {
>     font-family: 'Montserrat';
> }

The link you are using as font google is not working, I have another one : 
`https://fonts.googleapis.com/css?family=Montserrat&display=swap`

You are missing a `;` after the import so your margin: 0 and padding: 0 are not working;

It is always better to not use quote when selecting the font-family because it won't work on every browser.

for the green last section, your image is distored, instead of setting a max-height, try setting a max-width, the height will never go higher than the parent div (except if you decide so), if you set the max-width, the picture should never become distored.

Try also to set some margins on the section under the image1 section.

Keep it up and you will be there in no time, in the meanwhile if you have specific questions on my feedback, feel free to see me :)
PH

pascalbes:
Modifications réalisées, merci encore pour le coup de main !

Jepassaispar:
No probs, good job :)

PR #1581, opened 2020-01-07
[PAR-FT-012020] - María.MUNIZAGA - Ongoing

mmunizaga:


starter-code/css/main.css
```diff
@@ -1,8 +1,168 @@
...
+    align-content: center;
+}
+
+.text-box {
+    flex-direction: column;
+    justify-content: space-between;
+    align-content: center;
+    width: 50%;
+    text-align: left;
+    margin-bottom: 50%;
```
frankmarmier: The margin 50 % causes to have a white space at the bottom of your page, when you take it off it doesnt change anything to the main layout.

starter-code/css/main.css
```diff
@@ -1,8 +1,168 @@
...
+
+.green-box {
+    width: 100%;
+    height: 650px;
+    background-image:url("../images/spotify-icon-white.png");
+    background-position: center;
+    background-size: 100px;
+    background-repeat: no-repeat;
+    background-color: #00B172;
+    position: relative;
```
frankmarmier: You positioned your .green-box relatively, this causes your nav-bar, to appear behind the green-box element when you scroll down.
You can either remove this property because you don't position any element inside your green box absolutely, or you could increase the z-index on the nav-bar to be higher than the green box so it would appear infront of it.

starter-code/css/main.css
```diff
@@ -1,8 +1,168 @@
...
+    padding: 2px;
+    font-size: 1em;
+}
+
+.nav-bar div img {
+    width: 125px;
+}
+
+.conteiner {
+    background-image:url("./../images/landing.jpg");
```
frankmarmier: You could add the property: 
background-position: center;
so the image would be centered and not only the top part of the image would appear due to the fact that you fixed a height to 650px and that the image is a bigger than that.

starter-code/css/main.css
```diff
@@ -1,8 +1,168 @@
...
+    background-repeat: no-repeat;
+    background-size: cover;
+    background-color: transparent;
+    height: 650px;
```
frankmarmier: If you want the ".conteiner" to take the whole screen, as in a landing page, you could set the height to be 100vh, the "vh" unit being the size of the viewport, whether it's on mobile or desktop, it would be 100vh of the device's viewport.

https://developer.mozilla.org/en-US/docs/Learn/CSS/Building_blocks/Values_and_units

You could learn more about the css relative units in the relative section of this page.

starter-code/index.html
```diff
@@ -5,40 +5,76 @@
...
+        <article>
+          <h4>Discover</h4>
+          <p>Enjoy new music every Monday</p>
+          <p>with your own personal playlist.</p>
+          <p>Or sit back and enjoy Radio.</p>
+        </article>
+      </div>
+      <div class="img-box">
+        <img src="images/spotify-app.jpg" alt="app-img">
+      </div>
```
frankmarmier: Here the structure should change a little bit.

I put a better structure below, the reason for this, is that you can see that the box and the image on the left appear below the "yeezy" title.
It's a little detail but it changes the structure a little bit.

And this will of course change the css a little bit

 <section class="green-box">
      <h3>It’s as yeezy as Kanye West.</h3>
      <div class="wrapper">
        <div class="text-box">
          <article>
            <h4>Search</h4>
            <p>Know what you want to listen to?</p>
            <p>Just search and hit play.</p>
          </article>
          <article>
            <h4>Browse</h4>
            <p>Check out the latest charts,</p>
            <p>brand new releases and great</p>
            <p>playlists for right now.</p>
          </article>
          <article>
            <h4>Discover</h4>
            <p>Enjoy new music every Monday</p>
            <p>with your own personal playlist.</p>
            <p>Or sit back and enjoy Radio.</p>
          </article>
        </div>
        <div class="img-box">
          app-img
        </div>
      </div>
    </section>

starter-code/css/main.css
```diff
@@ -1,8 +1,168 @@
...
+}
+
+.service-box p {
+    font-weight: lighter;
+    font-size: 0.8em;
+}
+
+.green-box {
+    width: 100%;
+    height: 650px;
```
frankmarmier: I wouldn't set a fixed height to this green box, i would just let the elements inside make the box grow accordingly to the space they take, otherwise you would have to fix everything within in pixels in order to be able to fit in the 650px available space in height.
In addition if you remove the with 100%, it will fix the margin issue that you had.

starter-code/css/main.css
```diff
@@ -1,8 +1,168 @@
...
+    background-repeat: no-repeat;
+    background-color: #00B172;
+    position: relative;
+    box-sizing: border-box;
+    margin: 0 25px 0;
+    padding: 50px;
+    color: white;
+    display: flex;
+    justify-content: space-between;
+    align-content: center;
```
frankmarmier: You wouldn't need to display flex this element due to the layout change.

CF: the html section i changed.

starter-code/css/main.css
```diff
@@ -1,8 +1,168 @@
...
+    width: 50%;
+    text-align: left;
+    margin-bottom: 50%;
+}
+
+.text-box article p {
+    line-height: 0.6em;
+    font-weight: lighter;
+    font-size: 70%;
```
frankmarmier: It better to set the font-size in px, rem or em, a percentage might change according to the screen size / available space.

starter-code/css/main.css
```diff
@@ -1,8 +1,168 @@
...
+    color: white;
+    display: flex;
+    justify-content: space-between;
+    align-content: center;
+}
+
+.text-box {
+    flex-direction: column;
+    justify-content: space-between;
+    align-content: center;
```
frankmarmier: these properties don't work due to the fact that the element is not displayed flex.

frankmarmier:
Come see me so we can fix a little bit the layout together as i understand that there are alot of comments on this PR haha.

starter-code/index.html
```diff
@@ -5,40 +5,76 @@
...
+        <article>
+          <h4>Search</h4>
+          <p>Know what you want to listen to?</p>
+          <p>Just search and hit play.</p>
+        </article>
+        <article>
+          <h4>Browse</h4>
+          <p>Check out the latest charts,</p>
+          <p>brand new releases and great</p>
+          <p>playlists for right now.</p>
```
frankmarmier: Here you are creating multiple "p" to put them on below another, you should in theory only have 1 p and the text would wrap according to it's container.

starter-code/css/main.css
```diff
@@ -1,8 +1,168 @@
...
+}
+
+.icons {
+    display: flex;
+    justify-content: space-around;
+}
+
+.service-box {
+    width: 200px;
+    height: 200px;
```
frankmarmier: Having a fixed width and height here causes your h4 inside the service box to overlap because it takes more space that the fixed height you have defined.

frankmarmier:
good job !

PR #1582, opened 2020-01-07
[PAR-WDFT]-Kathleen D. ongoing

kathd:


ta-web-paris:
Nice hover on the links ! :) The only thing is that your last section is a bit too wide, but otherwise great job!

PR #1583, opened 2020-01-07
[PAR-WDFT]-MORSLI ongoing

PaulineLalou:
I don't know why the nav bar seems to have a padding when you scroll


ta-web-paris:
You need to override the default settings and put "margin: 0" in the body, then you won't see the padding on top. Also there are some small things to fix, like the Kanye image which is taking too much space in the box, and some titles have the wrong color :)

PR #1585, opened 2020-01-08
[PAR-WDFT] - DICKERSON ongoing

pauldkn:


ta-web-paris:
Hey Paul !
Your work looks really great :) Here are some small details:
- In your html I saw you used a br to make some space. Br has a semantic meaning: it should be used when there is a break in the text (ex: in a poem). To make a space visually, you could have added a margin to either your title or your paragraph :) It's a cleaner way to do it
- In your navbar, another way of doing what you did by giving the logo a 10% and the navlinks a 90%, you could have used justify-content:space-between on your header.But your way works too
- WHen you use flexbox, flex-direction: row is the default option, so you don't need yo specify it :) it will save you some lines of code
- Instead of adding a footer for the white space on the bottom of your screen, you could have add a margin to your green section
Great work :) keep going!

PR #1586, opened 2020-01-08
[PAR WDFT] Louis MD ongoing

louis-md:


ta-web-paris:
Hey Louis :) Your work looks great! Here are some advice for vert small improvements!
- flex-direction:row is the default value when you use display:flex. So you dont need to specify it. It will save you some lines of code :)
- instead of vertically aligning element with margins inside the navbar, you could use the align-items:center property inside your nav tag :)
- instead of fixing the height of the hero image with px you could use calc: https://www.w3schools.com/cssref/func_calc.asp . Your image will have a height of calc(100%-the height of your navbar)
- Instead of adding a margin-left to the app-image to make it on the right of the screen, you could use a justify-content:space-evenly on your section3-row class (and keeping a small margin on your image) and then your 2 elements in the class would be spaced correctly
Great work, keep going:)

PR #1589, opened 2020-01-08
[PAR-WEB] Camille ongoing

camilledlr:
It remain a little problem (among others) : I don't manage to remove the margin in the header section !


clarabesset:
Almost perfect! To remove the margin over the header you need to override the default properties of the body tag, you can see how it works it using the chrome console (screenshots below - in italic the default properties of the tags). The margin is not actually on the header but on the whole body, and since the header is the first block it appears to be on that element, but changing the properties of that element wouldn't work. Otherwise really good work, keep it up!!!

Capture d’écran 2020-01-09 à 09 28 41
Capture d’écran 2020-01-09 à 09 29 00

PR #1590, opened 2020-01-09
[PAR - WDFT] Remy Pham [ongoing]

kurosaek:


starter-code/css/main.css
```diff
@@ -1,8 +1,197 @@
...
+    margin-bottom: 40px;
+    margin-top: 40px;
+}
+
+
+.last-section p {
+    font-weight: lighter;
+    width: 200px;
+    text-align: left;
+}
```
frankmarmier: You might know it already, but commenting just in case.

These selectors are "dangerous" in a way that if you add an element to the last section that is not an article, these styles will be applied to it.
You could add a class to the article and access  the h3 through the article class.

starter-code/css/main.css
```diff
@@ -1,8 +1,197 @@
...
+
+/*   FIN SECTION 2   */
+
+
+.footer {
+    background-color: #00b172;
+    display: flex;
+    padding: 50px;
+    margin: 20px;
+    position: static;
```
frankmarmier: the default position of an element is static.

frankmarmier:
The clone looks good on a smaller viewport, responsiveness wasnt asked so i'm closing the PR.

Well done ! : )

PR #1593, opened 2020-01-09
WDFT Berlin Sanka Vaso

san4eto:
1st version


starter-code/css/main.css
```diff
@@ -6,3 +6,148 @@ Green: #00B172
...
+}
+.navitem div {
+  padding-left: 30px;
+}
+.logo {
+  height: 50px;
+}
+
+.headbanner {
+  background-image: url("/W1D4/lab/lab-css-spotify-clone/starter-code/images/landing.jpg");
```
DalinaLeila: this will not show up on any other computer but yours since the path you are giving to show image includes folders that only you have on your computer. Best is to write a `./` and then click enter depending on the path you want to have. This way it will not only be visible locally but also on any other computer.

starter-code/css/main.css
```diff
@@ -6,3 +6,148 @@ Green: #00B172
...
+  font-weight: lighter;
+}
+
+.greenbox {
+  display: flex;
+  flex-direction: row-reverse;
+  justify-content: space-between;
+  background-color: #00b172;
+  color: white;
+  background-image: url(/W1D4/lab/lab-css-spotify-clone/starter-code/images/spotify-icon-white.png);
```
DalinaLeila: see comment above.

starter-code/css/main.css
```diff
@@ -6,3 +6,148 @@ Green: #00B172
...
+  width: 370px;
+}
+.infoitems {
+  display: flex;
+  align-items: center;
+  width: 100%;
+  height: 800px;
+  background-size: cover;
+  flex-direction: row;
+  justify-content: flex-start;
```
DalinaLeila: why flex-start?

starter-code/css/main.css
```diff
@@ -6,3 +6,148 @@ Green: #00B172
...
+  font-size: 2.5em;
+  margin-top: 100px;
+  width: 370px;
+}
+.infoitems {
+  display: flex;
+  align-items: center;
+  width: 100%;
+  height: 800px;
+  background-size: cover;
```
DalinaLeila: why?

PR #1594, opened 2020-01-09
[BER-FT]Jason-Minatallah

jfrank325:


fenrew:
Nice work!
A few pointers:
- The Spotify logo has been clumped up into the corner in the nav bar, give a little space to breathe, it'll make it look much neater. Hint: Pandding 
- The main image can be handled a bit better as it gets cropped on a bit screen and the viewer will not be able to see the whole beauty of it. Hint Background-size: 100% 100%.
- The last part (green Kanye West box): I'm not sure what happened here if you run out of time or got a bit tired, but this part could need a bit more love. There's a bit too much green area here, so the size proportions between the background and the contents can be imporved significantly.

Otherwise a really good job!:)

PR #1595, opened 2020-01-09
[BER-FT-0120] Mia Lavanti & Johnny Burke

JohnnyAL:


starter-code/css/main.css
```diff
@@ -1,8 +1,181 @@
...
+}
+
+.navbar a {
+  padding: 20px;
+  text-decoration: none;
+  color: black;
+}
+
+header {
+  background-image: url(/lab-css-spotify-clone/starter-code/images/landing.jpg);
```
fenrew: Leaving a path like this should be avoided as it can easily cause errors like the one that I currently have when opening your project where the whole image is gone. I also believe (not 100% sure though) that this can in some cases cause security breaches as people can see where in the folders the different data has been stored.
In the future: Write all local urls like this: background-image: url("./images/landing.jpg"); and remember: ../ means go up one directory and ./ means go down one directory

PR #1596, opened 2020-01-09
[BER-FT-012020] - Jacqueline C. & Lluis M.

chenjacqueline:


fenrew:
Not gonna lie, this was pretty much spot(ify) on! Beautifully done!
It is not necessary to do it in this exercise, however when making web pages it can in several cases be quite advantageous to use id instead of class in your html. This will make searching for content a lot easier.
One other minor thing though, I a part of the image (main image) when you watch the page without the inspect element open. I would personally use background-size: 100% 100% instead of "cover" as this will keep the full image and not crop it in any ways.

But once again, this was very well done!

chenjacqueline:
Haha, love the pun! Ooh, great tips. Thank you, will keep in mind for the next lab. 🙌🏻

chenjacqueline:
> Not gonna lie, this was pretty much spot(ify) on! Beautifully done!
> It is not necessary to do it in this exercise, however when making web pages it can in several cases be quite advantageous to use id instead of class in your html. This will make searching for content a lot easier.
> One other minor thing though, I a part of the image (main image) when you watch the page without the inspect element open. I would personally use background-size: 100% 100% instead of "cover" as this will keep the full image and not crop it in any ways.
> 
> But once again, this was very well done!

PR #1602, opened 2020-01-09
[BER-FT] Marcel - Mohamed

Marsu-R:


starter-code/css/main.css
```diff
@@ -6,3 +6,189 @@ Green: #00B172
...
+
+.main-section {
+  margin-top: 70px;
+}
+
+.spotify-logo {
+  height: 100%;
+}
+
+h1 {
```
fonsogms: nice job guys!
Small note, avoid using h1 or h2 in css and use classes, identifiers and other type of selectors instead.

ciwan-aram: Thanks for feedback :)
No h1 and h2 in css ever again 👍

PR #1605, opened 2020-01-09
BE_FT_012020_Akua

akups:


starter-code/index.html
```diff
@@ -1,44 +1,121 @@
...
+       <nav class="navbar" >
+        <img class="imagelogo"  src="images\spotify-logo.png" alt="Logo" />
+        <ul class="linking">
+          <li><a href="#"></a>  Premium</li>
+          <li><a href="#"></a> Discover</li>
+          <li><a href="#"></a> Help</li>
+          <li><a href="#"></a> Download</li>
+        </ul>
+      </nav>
+    </div>
```
fonsogms: Nice job! 
In this case I wouldn´t have wrapped a nav in the div as it wasnt necessary. :)

starter-code/index.html
```diff
@@ -1,44 +1,121 @@
...
 
-    Millions of Songs
-    There are millions of
-    songs on Spotify
+    <div>
+      <header>
```
fonsogms: Same goes for the header, the div is not needed in this cases as tags such as nav, header where they already hold elements inside are not strictly necessary to be wrapped in a div

starter-code/css/main.css
```diff
@@ -6,3 +6,166 @@ Green: #00B172
 White: #FFF
 
 */
+@import url("https://fonts.googleapis.com/css?family=PT Sans");
+
+body {
+  font-family: "PT Sans";
+  border: 1px solid #f0f2f5;
+  margin: 10px;
+}
```
fonsogms: In this case I wouldn´t have use the body to style as I can see you had a lot of problems with margins and the borders

starter-code/css/main.css
```diff
@@ -6,3 +6,166 @@ Green: #00B172
...
+
+.what {
+  padding: 100px 200px;
+  display: flex;
+  flex-direction: row;
+  justify-content: space-evenly;
+  align-items: center;
+  height: 300px;
+  width: 1200px;
```
fonsogms: Following up with the previous review, the width is very wide so it makes it look that your website is a bit disordered. So maybe try taking out body and the width of `.what`see how it looks! :)

PR #1606, opened 2020-01-09
[BER-WDFT-012020] - Shahinya H. & Victoria B.

sha-hinya:
done

### DIFF
--- a/starter-code/css/main.css
+++ b/starter-code/css/main.css
@@ -1,8 +1,123 @@
-/*
-Colors:
+body {
+    font-family: ‘Lato’, sans-serif;
+}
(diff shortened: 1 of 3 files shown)


starter-code/css/main.css
```diff
@@ -1,8 +1,123 @@
...
+    justify-content: baseline;
+    align-content: center;
+}
+
+.perks img {
+    width: 150px;
+    height: 120px;
+}
+
+hr {
```
fonsogms: Avoid using the html tag selector, and always try and using classes or identifiers instead as it is best practice that way :)

PR #1621, opened 2020-01-10
[PAR-WDFT-0120] - Yannick Bourenane ongoing

yannick-bourenane:
Just few changes after your feedback (font-family + topbar)

I didn't merge the #topbar and .wrap div cause I needed the blank background to be a 100%

But maybe there is a more efficient way of doing it.

Thanks !


TatijanaRajcic:
Great job :)

PR #1624, opened 2020-01-10
done [PAR - WDFT] - Shannon K ongoing

tranquildoe:


clarabesset:
Good job! I see you understand flexbox and as I said it's no big deal that your last section isn't perfect :)

Just a couple little mistakes on lines 29 and 43, make sure you close your tags using a backslash "</p>". Let me know if your VScode doesn't underline these end tags because that'd be odd, it's supposed to let you know about such syntax errors.  

On lines 40, 47, 54 you are using the same id "one" multiples times, but you can only do that with classes (more here: [https://css-tricks.com/the-difference-between-id-and-class/]). IDs are unique, meaning you can only target one element with them, that you need to have special properties for. Classes are for when you want to reapply a style to multiple elements :)

Line 49, you wrote "id=#adjust", but you can only use the # in the CSS file. You wrote it correctly in other places so I'm assuming you knew! --> " id="adjust" " in the HTML (like for classes).

Last one, you're not closing your end tag for the very last image. That's just a typo but it can mess with your whole page, watch out! 

Come to me if you're not seeing all of this in VScode, that could mean the parameters are set right :)

PR #1626, opened 2020-01-10
PAR-FT-012020-Youval BUGAJSKI ongoing

YuvalBu:


TatijanaRajcic:
Hey Youval! Your page looks really great. Here are some advice to make it even better:

- You should not use the "br" tag to separate paragraph. The "br" tag has a semantic meaning (it represents a break for example in a poem). Instead, try to use "p" tag, and play on their margin
- In your CSS: when you use "display:flex", the default value is "flex-direction: row". so you don't need to specify it. You need to specify it when you change the value, using the column for example: "flex-direction:column"
- Instead of using padding top / bottom to center your nav elements vertically, it's better to apply on your parent element (header) "align-items:center". This is a property you can use since you are using flex. Come and see me if you are not clear with this property :)
- You should have used "position:fixed" for the navbar. It's just a small change in your code. You need to set the position to fixed, and then use the property top and set it to 0, to actually tell your code where you want to fix it.
- Using vh is useful when you want a part of your page to be an precise size. This is useful for hero images for example (since we want them to generally take the whole screen). But in the case of your section "about", if you remove the "height" you'll see that it's better sized. If you want to add some blank space around it, better use padding
- Personnally, I wouldn't call classes and ids the same (as you did with kanye-west) as it can be confusing in the end when you'll have a lot of them. But that's a personal advice :) 
- There is a way to simplify your kany-west section. You don't need to apply display:flex on your class "kanye-west". What you did on your div with id="kanye-west" is enough. Therefore: you'll only have this code:
```
.kanye-west {
  height: 100vh;
  background-color: white;
  padding: 2%
}

#kanye-west {
  display: flex;
  flex-direction: row;
  justify-content: space-evenly;
  align-items: center;
  text-align: center;
  background-color: #00B172;
  height: 96vh;
}
```
- You shouldn't use "hr" to underline a text. "hr" has a semantic meaning, it's not only for styling. Instead, use the border property.

Come and see me if you want us to review how you could have simplified the last section.
Good job, keep going :))

YuvalBu:
Hi Tatijana,

Thank you for all your help.

Yuval

PR #1632, opened 2020-01-14
done [PARWDFT Andrew NB] ongoing

andrewnunezbibby:


TatijanaRajcic:
Hey Andrew! Good job :) Here are some of my remarks:

- For the logo, you used background-image. It works here but I wouldn't see it's the most common to use for an logo image. What you could also do is add the image in your HTML (with the img tag) and then for your CSS have this class:

// HTML:
```
  <header>
    <img id="logo" src="./images/spotify-logo.png" alt="">

    <ul class="nav-bar-ul menu"
        style="list-style-type:none;">
      <li>Premium</li>
      <li>Discover</li>
      <li>Help</li>
      <li>Download</li>
    </ul>
  </header>
```
// CSS
```
header #logo {
    height: 100%;
    width: auto;
}
```
- You don't need to specify "flex-direction: row" when you use flex. This is the default value.

- To align elements vertically, you do not need margins/paddings. Since you applied flex, you could use align-items: center to center elements vertically (if your flex-direction is row).

I simplified your code (with comments):
```

header {
    position: fixed;
    height: 80px;
    width: 100%;
    display: flex;
    align-items: center; /* property you use to center the logo and the ul vertically */
    z-index: 1;
    background-color:#FFF;
    padding: 10px; /* you add a padding to the navbar so you don't need to add it the all its elements */
    box-sizing: border-box;
}

header #logo {
    height: 100%;
    width: auto;
}

header h1 {
    font-size: 5em;
}

header .menu {
    display: flex;
    justify-content: space-around;
    align-items: center; /* you center the elements inside the ul */
    font-size: 0;
    width: 500px;
}
```

- To make sure your hero image takes the whole screen, instead of using pixels, you could use calc:
```
#jumbotron {
    height: calc(100vh - 80px); // HERE
    padding-top: 50px;
    background-position: 0, 0;
    background-repeat: no-repeat;
    background-size: cover;
    position: relative;
    color: white;
    text-align: center;
  }
```
This is a CSS property that does caculation. Here, what I'm doing is basically setting the height of the hero image to 100vh (total height of the page) - the height of the navbar.

- You could have added a margin to your #functionalities so you would have a space around your green container as in the model

- To underline elements as in the model, you should style the border (by adding a border bottom)

- Be careful of the classes you add which are not useful. For example, ".function" does not change anything.

Overall, I see that you know how to use flexbox so it's great :) don't hesitate if you have any question!

PR #1661, opened 2020-01-15
done

MichielSnoek:


starter-code/css/main.css
```diff
@@ -6,3 +6,135 @@ Green: #00B172
...
+.green-box h4{
+    font-weight: bold;
+    font-size: 2em;
+}
+
+.green-box p{
+    font-size: 1.3em;
+}
+
+.green-box .spotify-app {
```
LloydC: I think this class isn't necessary if you only want to apply a fixed width, you can use "" directly.

PR #1690, opened 2020-01-27
[LIS JAN 20 Raquel Coelho] 

RaqueCoelho:


soniameller:
Thanks Raquel! 🙃
You don't need to type the name of the lab. Just campus, squad and your name.
**[LIS JAN 20 - Raquel Coelho]**

PR #1702, opened 2020-02-06
[BER-WDPT-022020]) - Sonja L.

Asphalthexe:
what I did in class

### DIFF
--- a/starter-code/index.html
+++ b/starter-code/index.html
@@ -1,70 +1,82 @@
 <!DOCTYPE html>
 <html lang="en">
-  <head>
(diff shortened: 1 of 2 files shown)


starter-code/index.html
```diff
@@ -40,5 +63,8 @@
     Enjoy new music every Monday
     with your own personal playlist.
     Or sit back and enjoy Radio.
+  
+  </div>
+
   </body>
 </html>
```
numbing: cool, just please try to finish it later when you have time but so far so good

PR #1704, opened 2020-02-06
BER WDPT Ian & Nomu

ian-kerins:


starter-code/index.html
```diff
@@ -5,40 +5,87 @@
...
+
+    
+    
+
+    
+    
+
+    
+    
+
```
numbing: nice job, keep going

PR #1736, opened 2020-02-19
LIS - PTWD - FEB - Rui Pereira

RuiMPer:
tag @ironhack-LIS-PTWD
having difficulties organizing the html
css:
difficulties to separate the nav from the image.

### DIFF
--- a/starter-code/css/main.css
+++ b/starter-code/css/main.css
@@ -7,7 +7,6 @@ White: #FFF
 
 */
 nav {
-	position: fixed;
 	width: 100%;
 	display: flex;
 	flex-direction: row;
@@ -15,17 +14,20 @@ nav {
 	background-color: white;
 }
 .logo {
-	height: 3em;
-	padding: 1em;
+	height: 3.5em;
+	padding: 1.25em;
 }
 
 .buttons-nav li {
+	font-size: 1.2em;
+	text-align: center;
 	display: inline;
 	padding: 1em;
 	list-style-type: none;
 	margin-left: 1em;
 }
 .buttons-nav li a {
+	font-size: 1.25em;
 	color: black;
 	text-decoration: none;
 }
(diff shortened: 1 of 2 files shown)


ironhack-LIS-PTWD:
Hi Rui,

As you asked I am currently having a look at your Spotify lab. 
I can see that here again your css file was not linked correctly. 
Change it to this instead:  ` <link rel="stylesheet" href="/css/main.css" />`
Same thing with your images, this is what you should have in the `src` 
`src="/images/spotify-logo.png"`.

You are using the `<section></section>` tags appropriately here.
Maybe instead of the `<article></article>` tag I would be using a `<div></div>` for the three icon blocks since they are just for layout and visual separation but they belong to the same "section" of the website.

This is a very good reference to guide you if you have any doubts:
h5d-sectioning-flowchart
[source here]( http://html5doctor.com/lets-talk-about-semantics/)

Something else important to keep in mind is the difference between the `head` and the `body`
I can see that you had the `nav` tag inside of your head instead of body and you definitely don't want to do that.

The `head` holds everything that are links to other files (stylesheets or fonts for example) and meta data (such as viewport setting).
whilst the `body` holds the tags that you actually want to display in the website such as the
`nav` `h2` and `p` etc.. 
The only thing you would be adding in the body that is not "displayed"  are scripts (linking javascript files).

Overall try to keep things simple. and look at woking examples using the google chrome tools. Overthinking your html will only cause more problems than it’s worth. 

Regarding your CSS 
It looks like you are mixing up `float` and flex box a little bit. I would recommend you not to use float and only use flex box. A float property will be ignored in a flex container.

I hope this helps, if you have any more specific questions let me know :) 

Paula

ironhack-LIS-PTWD:
I also just added the [solution code](https://github.com/ironhack-LIS-PTWD/lab-css-spotify-clone) to the learning unit so you can compare the code

PR #1770, opened 2020-02-26
LIS - WDPT - FEB - Julia

juliamendesc:
@ironhack-LIS-PTWD

After burning some neurons, I finally managed to deliver something which I think is decent. However, I had some difficulties with flexbox in regards to the green section (it is not as spaced between than I would like it to be) and in the navbar (same reason).

Could you shed some light on this for me? 
Thank you!

### DIFF
--- a/starter-code/css/main.css
+++ b/starter-code/css/main.css
@@ -1,3 +1,147 @@
+html {
+    font-family: 'Montserrat', sans-serif;
+    font-size: 12px;
+}
+
+.nav-bar {
+    margin-right: auto;
+    padding: 5px 10px;
+    display: flex;
+    justify-content: flex-end;
+    flex-direction: row;
+}
+
+.head-container {
+    display: flex;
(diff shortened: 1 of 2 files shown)


ironhack-LIS-PTWD:
Hey, well done, Julia! It's looking very good.

For the final section, I recommend giving your` .green-flextext`class a width and padding for the spacing. 
```
.green-flextext {
    text-align: left;
    width: 40%;
    padding: 30px;
}
```
This should help :) 

For the navbar, instead of using <ul></ul>, you could make each text element a link within the <nav> </nav>
tags for example: 
```
<nav>
      <a href="">Premium</a>
      <a href="">Discover</a>
      <a href="">Help</a>
      <a href="">Download</a>
    </nav>
```
and then, you can give some CSS styling to `nav>a`  including some padding. 
That way each element can be spaced out.

Let me know if you need any more help :)
Lissy

PR #1791, opened 2020-03-24
PAR WDFT Divya - Spotify Lab ongoing

dgopal23:
Hello, 
Here is my spotify clone exercise. Please let me know how I can improve!

Thank you very much,
Divya


dgopal23:
Hello,
I managed to fix my path links ! I hope it's okay!
Thank you!

ta-web-paris:
Very good ! 
A few points :
- the "flex direction: row" is the flex behavior by default, so it should work without having to use it ! 
- now that you're a pro with css and flex, i recommend you to look into using "%" or "vh / vw" for sizing your divs as it might be more flexible on different screen sizes, you can find a lot of information online and you will have to use it in the future
- you can use "align items : baseline // or // flex-start" for your arguments section, so the titles are aligned properly
- font sizes / weight small adjustments (the paragraph in the hero landing is a tiiiiny bit too big) 
- the white logo in the green section : you can use it as a background image instead of an html element, since it's easier to maintain if you have more content in the green section (that would overlap on the white logo)

PR #1793, opened 2020-03-24
PAR WDFT 032020 Sam L ongoing

samlfair:


ta-web-paris:
Very nice job, super clean 
I love the sticky header with the shadow !

The only thing I could say is
- in your header, you have three containers before the logo and nav-list, why so much haha ? You can drop some and have shorter code.

I think you're very comfortable with CSS, I challenge you to do the Instagram clone in the bonus section of the week ! It's nice practice and one advice for later : you can create a "flex" class so you don't have to repeat your flex declarations everywhere, (even flex-column, flex-center etc... treat yourself !)

PR #1794, opened 2020-03-24
[PAR WDFT 032020] William Hermozo ongoing

williamhzo:


TatijanaRajcic:
Hey William! Thanks for your really nice work! I don't have much to say but here are some remarks to make it even better:
- in your navbar ul, make the "li" the direct children of the ul, and then put the anchor inside the li instead. Otherwise it's not semantically correct.
- just one quick thing: "alt" attribute on images are mainly used for accessibility. This is the text that is going to be read in case there is an issue with displaying the image, or by screen readers for people with a visual impairment. To make it more clearer, I would recommend you put a small text, instead of using kebab case as in classes.
Very nice work!

PR #1795, opened 2020-03-24
[PAR-WDTF] – Rebecca Norén 🍕 ongoing

rebeccanoren:
Creating HTML structure and added styling.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+  "liveServer.settings.port": 5501
+}
--- a/index.html
+++ b/index.html
@@ -1,18 +1,86 @@
 <!DOCTYPE html>
 <html lang="en">
-  <head>
-    <meta charset="UTF-8" />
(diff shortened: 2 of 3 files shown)


TatijanaRajcic:
Hi Rebecca! Thanks for your nice work! Here are some tips to make it even better:

- I would advise you to use Ul list for the Navbar links as it is the norm. You can find [here](https://stackoverflow.com/questions/36811224/why-do-navigation-bars-in-html5-as-lists) people debating over the use of Ul for Navbar links, and as you can see, it wins :D 

- while using display:flex, you don't have to specify "flex-direction: row" as it is the default value. Same goes with "flex-wrap: nowrap;". In your "left-navigation" class, the property "justify-content: flex-end; " does not do anything either here

- you can simplify your "what is on spotify" class. you don't need "justify-content:flex-start" (because it's the default value), and you don't need align-content: center neither.
- same in ".our-offers" you don't need the default values.

Well done Rebecca!

PR #1797, opened 2020-03-24
[PAR WDFT] Guillaume de Glisezinski ongoing

TheShua:


ta-web-paris:
Hey good job !
A few points :
- You don't need to use "line height" to size the height of elements, "height" does the trick 
- for the white logo in the green section, you can set it as a background image for the whole section, no repeat, position center center and that way you don't have to bother with calc

Nice work anyway :)

PR #1800, opened 2020-03-24
[PAR - WDFT - 0320] Florian Jomain ongoing

florian-jomain:


ta-web-paris:
Nice and clean !

A few points :
- you can use "border-bottom" to do the green line under the title in the white section (if the line is taking all the space, you can use the "display: inline-block" to make the box only the size of the title)
>> that way you can style the line how you want it to be : the space between the text and the line, thickness, color, if it's dots or a double line, or a plain line... anything you want !

- you can look into vh / vw for sizes of your sections (it's like a percentage but depending on your browser's window size)

Nice use of selectors and flex boxes ! The result is very good
If you have questions don't hesitate ! 

Nina

PR #1802, opened 2020-03-24
PAR WDFT 0320 Guillaume Marcellin ongoing

guiohm432:


ta-web-paris:
Hey Guillaume ! 
It seems you pushed an old version of the lab (the files have changed a liiiiiitle bit since) so try to fork again and start from scratch next time since it's a) good practice b) for this one it's fine but maybe the next labs changed a bit and you need to make sure you're working on the right files :)

Your HTML has a few things that could be corrected : 
- try not to leave too much space between tags (between your "aside" and "header" for example) it's ok if it's to separate sections visually but here it's just confusing (correct syntactically don't worry)
- the "aside" tag should be a "p" maybe, since it's right under the title, it's more considered as a paragraph (an "aside" would be in a weird place like the quote in the IronSkyDive prework or in a column next to the main content)
- your "footer div" and "footer text" could be only one div, since they're making a box in a box, it's easier to work on afterwards if you simplify :)

otherwise all good for HTML
for CSS a few points as well : 
- the nav bar can have no border or maybe just a bottom border to be cleaner 
- your spotify logo is a bit stretched, one way to fix that is to only change one of the properties (like the width for example) and it will keep the aspect ratio (tell me if you have trouble doing that, sometimes img are tricky to resize)
- you don't need to set a width on the list in the nav, if you have to put new elements it could become messy (here it's fine but the goal is to pick good habits for future projects !)
- try to assign the font-family in the body style, that way it's the same for the whole document, and for web design i recommend "sans serif" fonts (like Arial, Roboto, Helvetica...) as they are easy to the eye (except if you want a typewriter or a paper aspect)
- you should try and set font size with a new unit : em / rem (look online it's very easy to use ! 1 rem = the basic size of text (18px) and 1em = the basic size of the parent element). you can use that unit for margin and padding as well ! (you don't have to, but it's nice for responsive design, and when you want the margin and padding to be related to the text size)
- careful with the vh : at one point it's not working (because you put it with a "%" and on the "width" property) vh = vue Height (you should use vw = vue Width for the width property, it's weird but it's the rule haha) and never use % AND vh or vw, because it's only 1 unit per value
- for "all functionalities" : you should put "align-items : flex start // or // baseline" to have them well aligned vertically

it's quite a lot of remarks but your work is nice ! very good use of flex boxes, this is just a list of small details for you to improve on future projects (if you rework your spotify lab you can see the difference it will make and the whole thing will be perfectly clean)

Keep up and I'm happy to see you again :) :)  WELCOME BAACK
Please tell me if you have any question / I can help with something

Nina

PR #1804, opened 2020-03-24
PAR WDFT March Vladimir Leloup - needs imrpovement

Vlp90:
Exercice done

### DIFF
--- a/index.html
+++ b/index.html
@@ -5,14 +5,102 @@
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
     <meta http-equiv="X-UA-Compatible" content="ie=edge" />
     <title>Spotify Clone</title>
+    <link rel="stylesheet" href="styles/style.css">
+    <link href="https://fonts.googleapis.com/css?family=Montserrat&display=swap" rel="stylesheet">
     <!-- don't forget to link your styles -->
   </head>
+
+
   <body>
-    Premium Discover Help Download Music for everyone. Spotify is now free on mobile, tablet and computer. Listen to the
-    right music, wherever you are. What’s on Spotify? Millions of Songs There are millions of songs on Spotify HD Music
-    Listen to music as if you were listening live Stream Everywhere Stream music on your smartphone, tablet or computer
-    It’s as yeezy as Kanye West. Search Know what you want to listen to? Just search and hit play. Browse Check out the
-    latest charts, brand new releases and great playlists for right now. Discover Enjoy new music every Monday with your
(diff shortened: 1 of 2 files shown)


TatijanaRajcic:
Hello Vladimir! 
Well done for the exercice! Some remarks:
- I would advise you to add "padding:0" in your universal selector, because some browsers have default paddings too (not only margins)
- You can apply your font to the whole body directly, since you are using it everywhere. it will save you some words of code
- since you gave a class to your header, use it instead of using the generic tag :)
- IMPORTANT: your indentation is kind of weird so don't hesitate to check out the options that Prettier (the extension you installed with VSCode) has, they have automatic indentation if you don't want to worry about it yourself. It's very important because otherwise people will read your code and they will stop because it's not easily readible.
- There is some repetition of code between what you wrote for the tag "nav" and the properties you gave to the Ul inside.I simplified your header code to get rid of repetition:

In the CSS:
/* HEADER */
```
header {
    width: 100%;
    box-sizing: border-box;
    position: fixed;
}

.logo-img {
    width: 150px;
}

.flex {
    display: flex;
    align-items:center; 
}

.pre-header {
    padding : 15px 30px;
    height: 7%;
    background-color: white;
}

.pre-header ul {
    list-style:none;
    margin:0;
    padding:0;
}

nav a {
    text-decoration: none;
    color: black;
    margin-left: 20px; 
}
```

In the HTML:

```
<header class="pre-header flex">
  <img class ="logo-img" src="/images/spotify-logo.png" alt="spotify-logo">
  <nav>
    <ul class="flex">
      <li><a href="#">Premium</a></li>
      <li><a href="#">Discover</a></li>
      <li><a href="#">Help</a></li>
      <li><a href="#">Download</a></li>
    </ul>
  </nav>
</header>
```
- Please be careful to use " " around your link, when you use the property background-url, because it might not work on other computers, so avoid the risk :)

- For the "What's on Spotify" underling, the idea was not to add a "hr" (as this tag has a semantic meaning by itself. Here, you just want to undelrine a title. So it's just styling, no semantics. Use border-bottom for that (and transform the title into a inline-block element, so that the underlying does go the whole width)

- "flex-direction:row" is the default value, so you don't have to specify it.

- You don't have to give specific IDs for the purpose of styling just one property. For example, in your section 4, you could have target your h3, by using specific selectors. Instead of:

```
#padding-h3{
    padding-top: 40px;
}
```

.. you could have written it like that: 
```
.part-1 .padding:nth-child(3) {
    padding-top: 40px;
}
```


Try and change some of those things, and then send me a message :)

Otherwise, good work !!

PR #1809, opened 2020-03-25
PAR WDFT 0320 Teddy ROGERS needs improvement

TeddyRod:


ta-web-paris:
Hey Teddy :)

I know you said you had trouble with CSS. Don't worry, now it's just about practice (it will not get harder, just new techniques to learn to do cool stuff) so I'll give you some points to correct. I need you to work on them and push your changes, and tell me when you're finished ! (but please focus on the JavaScript labs, this one you have until next monday and it's not the priority)

HTML side : clean code, just a few points :
• do not use a div around the `<ul>` in the navbar. you don't even use the menuLink class in your css, so less code = less bugs ! you can remove the div without any trouble.
• same thing in your white section, don't use a div around the `<h3>`, and don't use the `<hr>` because we will use borders to make the color line, it's better and cleaner !
• same thing in your green section, remove the `<hr>` after the `<h3>`

CSS side : nice use of flexbox ! that's very important for your skills, so good job :)
• remove the "background color white" everywhere except your .nav, it's been bugging your green section (white is the default color, you don't need to say it unless you want to override a colored element, like your nav)
• your `<h3>` should have a border-bottom to make the line, green or white depending on the section. to style your h3 you should always go from the big section to the element; like
>> .whatIs h3{
     display: inline-block;
     bottom-border: 3px solid #00B172
}

>> .greenSec{
     background-color: #00B172;
     color: white
}

>> .greenSec h3{
     display: inline-block;
     bottom-border: 3px solid white
}

• try to style the big sections first, applying global flex rules and color themes, and then style the small elements for sizing, borders, paddings etc... you'll be cleaner, faster and you'll find exactly what's left to do

• you can't put flex onto `<h3>` or `<hr>` because they're elements inside a box, and flex is for the box itself ! 

I'll send you an example by Slack, please get back to me for ANY question, I know that this message is long but we can talk about it on Zoom it'll be easier to show you code / explain live :)

Nina

PR #1811, opened 2020-03-25
[PARIS-FT-032020]) - Yuval Bugajski ongoing

YuvalBu:


TatijanaRajcic:
Hey Yuval, thanks for your lab! It looks really nice! 

Please keep in mind that for the labs you should name their "PAR.. " or "PAR - .." and not "PARWD" or "PARIS" otherwise we don't find them while doing our filter. 

Here are some remarks on the lab:
- you use "justify-content:space-between" on the "li" but actually it's the "ul" that is in display flex, so you should right this property at the "ul" level. I just removed it from the "li" and it does not change anything for the design of your page.
- .what-is-spotify h2 and .underline-green  amount to the same element, so you could put all the properties in one, since you are not reusing the .underline-green class anywhere.
- if you wanted to make lines inside your paragraph without using spans everywhere, you could have given a fixed width to the container of your text, so that the text would go to the line. That being said, you would have had less control on which word marks the new line.

Good job :)))

PR #1815, opened 2020-03-26
BER - FTWD - Manuel - Benjamin 

Benchberlin:


styles/style.css
```diff
@@ -6,3 +6,153 @@ Green: #00B172
...
+    display: flex;
+    justify-content: space-around;
+    flex-direction: column;
+    color: white;
+    padding: 20px 20px;
+}
+
+#icon {
+    height: 80px;
+}
```
Jeff-Moraes: For this icon, you could also have added it as a `background-image` to the `actions ` section. 

`background: url("../images/spotify-icon-white.png") no-repeat; background-position: center; background-color: #00b172; background-size: 130px;
`

PR #1816, opened 2020-03-26
BER-FTWD-Nate-Sami

Alba-Tros:


styles/style.css
```diff
@@ -6,3 +6,157 @@ Green: #00B172
...
+
+#section4 p {
+    width: 35%;
+    font-size: 1.1rem;
+}
+
+#spotify-logo-white {
+    width: 8%;
+    position: absolute;
+}
```
Jeff-Moraes: nice approach! Another way do that:
You could also have added it as a `background-image` to the `section4` section.

background: url("../images/spotify-icon-white.png") no-repeat; background-position: center; background-color: #00b172; background-size: 130px;

styles/style.css
```diff
@@ -6,3 +6,157 @@ Green: #00B172
 White: #FFF
 
 */
+
+:root {
+    --custoTxtCol: 1a1a1a;
+    --custoGreen: #00b172;
+    --custoWhite: #fff;
+    width: 60%;
```
Jeff-Moraes: I don't think you actually need it, you could have just given a width 100%;

PR #1817, opened 2020-03-26
BER-FTWD-Aleah-Victoria

victoria-toria:


styles/style.css
```diff
@@ -6,3 +6,101 @@ Green: #00B172
...
+   
+
+h3 {
+    color:#00B172;
+    font-weight: bold;
+    //text-align: center;
+    font-size: 24px;
+}
+
+h4 {
```
fenrew: There is a lot of space between your text under "What's on Spotify". The reason for this is because the h3 and h4 that you are using have a lot of default margin. To reduce this space you can say:
h3, h4 {
 margin: 0
}

PR #1818, opened 2020-03-26
BER-FTWD-[Lucas Karl]

karlkristopher:


styles/style.css
```diff
@@ -6,3 +7,180 @@ Green: #00B172
...
+    padding: 0 1rem;
+    font-size: 1.1rem;
+
+}
+
+
+header {
+    background-image: url("../images/landing.jpg");
+    background-repeat: no-repeat;
+    background-size: cover;
```
fenrew: You are missing a lof of the background image, a way to at least show the nice part and not just the top of the image is to center it:

background-position: center;

styles/style.css
```diff
@@ -6,3 +7,180 @@ Green: #00B172
...
+    height: 500px;
```
fenrew: Having the height set in % or in vh is usually a lot better as it will make your website much more responsive

styles/style.css
```diff
@@ -6,3 +7,180 @@ Green: #00B172
...
+    display: flex;
+    flex-direction: column;
+    justify-content: center;
+	align-items: center;
+}
+
+h1 {
+    margin-top: -75px;
```
fenrew: Negative margins can be handy, but keep in mind that it can also screw up the code as well since it will put the element on top of whatever comes above it. I would in this case have set the margin: 0, however using it like you do here is perfectly fine:)

styles/style.css
```diff
@@ -6,3 +7,180 @@ Green: #00B172
...
+
+.kanyeimage {
+    width: 180px;
+    margin-right: 50px;
+    margin-top: 20px;
+}
+
+
+
+
```
fenrew: There is a lot of space between your text under "What's on Spotify". The reason for this is because the h3 and h4 that you are using have a lot of default margin. To reduce this space you can write:
p, h4 {
margin: 0
}

PR #1819, opened 2020-03-26
BER-FTWD:Yung-Ting+Yosra

yosrayousfi:
@yung-ting


styles/style.css
```diff
@@ -6,3 +6,131 @@ Green: #00B172
...
+  text-align: center;
+  font-size: 3.5rem;
+  margin-top: 0;
+  letter-spacing: 1px;
+}
+.landing {
+  align-content: center;
+  background-image: url(../images/landing.jpg);
+  background-size: cover;
+  background-repeat: no-repeat;
```
fenrew: There's a lot of the background image that is missing. I can recomment you to use:
background-position: center;

This will center the background image instead of showing the top part.

styles/style.css
```diff
@@ -6,3 +6,131 @@ Green: #00B172
...
+  height: 600px;
+}
+.landingText {
+  align-content: center;
+  color: white;
+  padding-top: 200px;
```
fenrew: This here is a bit hard-coding;)

You can instead display:flex on you landing class. And then you can center it with align-content: center and justfiy-content: center

styles/style.css
```diff
@@ -6,3 +6,131 @@ Green: #00B172
...
+
+.landing p {
+  font-size: 20px;
+  font-weight: 0;
+  letter-spacing: 1px;
+}
+.service {
+  display: flex;
+  flex-wrap: wrap;
+  justify-content: space-between;
```
fenrew: I would personally have used space-around here since it currently pushed all the way to to sides, making the design look a bit off:)
You can also fix the issue with setting a width of the container

PR #1820, opened 2020-03-26
BER-FTWD-Paula&Olha

olhadanylchenko:
done

### DIFF
--- a/styles/style.css
+++ b/styles/style.css
@@ -6,3 +6,150 @@ Green: #00B172
 White: #FFF
 
 */
+
+* {
+  box-sizing: border-box;
+}
+body {
+  font-family: Circular, spotify-circular, Helvetica, Arial, sans-serif;
+  margin: 0;
+  padding: 0;
+  width: 100vw;
+}
+
+.navbar {
+  width: 100vw;
+  height: 4.3rem;
+  position: fixed;
+  display: flex;
+  flex-direction: row;
+  justify-content: space-between;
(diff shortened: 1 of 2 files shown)


styles/style.css
```diff
@@ -6,3 +6,150 @@ Green: #00B172
...
 */
+
+* {
+  box-sizing: border-box;
+}
+body {
+  font-family: Circular, spotify-circular, Helvetica, Arial, sans-serif;
+  margin: 0;
+  padding: 0;
+  width: 100vw;
```
fenrew: The width here set to 100vw makes your website horizontally scrollable on some web browser. It is thus better to use 100% instead of 100vw, to prevent this from happening.

styles/style.css
```diff
@@ -6,3 +6,150 @@ Green: #00B172
...
+}
+
+.navbar {
+  width: 100vw;
```
fenrew: Same here

styles/style.css
```diff
@@ -6,3 +6,150 @@ Green: #00B172
...
+  list-style-type: none;
+  display: flex;
+  flex-direction: row;
+  align-items: center;
+  justify-content: space-between;
+}
+
+.section2 {
+  background-image: url("../images/landing.jpg");
+  background-size: cover;
```
fenrew: You can also put a   background-position: center;
This will center your image and make it look a bit nicer:)

PR #1821, opened 2020-03-26
BER-FTWD-Robert-Marie

bl00dymarie:
We tried our very best :); @robertbozsik

### DIFF
--- a/index.html
+++ b/index.html
@@ -5,14 +5,104 @@
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
     <meta http-equiv="X-UA-Compatible" content="ie=edge" />
     <title>Spotify Clone</title>
-    <!-- don't forget to link your styles -->
+    <link rel="stylesheet" href="styles/style.css">
+    
   </head>
   <body>
-    Premium Discover Help Download Music for everyone. Spotify is now free on mobile, tablet and computer. Listen to the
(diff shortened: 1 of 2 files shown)


styles/style.css
```diff
@@ -6,3 +6,117 @@ Green: #00B172
...
+}
+
+header img {
+    width: 150px;
+}
+
+.head {
+    margin: 5rem;
+    display: flex;
+    justify-content: space-around;
```
fonsogms: nice work! In here I would suggest to use space-between. :)

styles/style.css
```diff
@@ -6,3 +6,117 @@ Green: #00B172
...
+    margin: 0 auto;
+    text-align: center;
+    width: 300px;
+    border-bottom: 2px solid #00B172;
+}
+
+.icons {
+    text-align: center;
+    display: flex;
+    margin-top: 50px;
```
fonsogms: here I also would suggest to use justify-content:space-around, to spread the elements evenly

index.html
```diff
@@ -5,14 +5,104 @@
...
+          </h3>
+          <p>
+            Stream music on your smartphone, tablet or computer
+          </p>
+        </box>
+      </div>
+    </section>
+    
+    <section class="kanye">
+      <img src="/images/spotify-icon-white.png"></img>
```
fonsogms: Nice work! in here tho i would suggest to put the text in a div and then use flex box with the div, the logo and the image of kanye west, If you have questions regarding this let me know

PR #1824, opened 2020-03-26
BER-FTWD-Marcia-Dioni

marciafurtado:
Done

### DIFF
--- a/styles/style.css
+++ b/styles/style.css
@@ -1,8 +1,143 @@
-/*
-Colors:
+* {
+    font-family: Circular, Helvetica, Arial, sans-serif;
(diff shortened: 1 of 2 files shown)


styles/style.css
```diff
@@ -1,8 +1,143 @@
...
+  h1 {
+    font-size: 3em;
+  }
+  
+  .section3 {
+    text-align: center;
+    display: flex;
+    flex-direction: column;
+    justify-content: space-between;
+    align-items: center;
```
fonsogms: Here in the margin top you only need to put 10px; if not it will not work! :)

PR #1831, opened 2020-03-27
PAR WDFT 0320 Mathias - ongoing

saitam1990:


ta-web-paris:
Hey Mathias :)

Nice job !! I'm going to make a big review to give you the best advices for future practice, but you don't need to send me a correction I validate your lab ! It's just for you if you want to work on it on your free time or to keep in mind for next time :)

HTML : very nice use of flex classes !! That's a good thing to do since you don't repeat yourself writing "display flex" everywhere in the css, good reflex ! But : 
- Dans ta barre de navigation en haut, tu peux utiliser des tags plus spécifiques : NAV au lieu de DIV pour ta navigation, et UL + LI (avec des A dans les LI) au lieu de DIV + H2 pour tes liens en haut à droite (puisque c'est une liste de liens). Le HEADER, c'est au choix, soit tu enlèves le DIV / NAV et tu mets IMG et UL direct dans HEADER, soit tu enlèves HEADER et tu gardes NAV avec tout dedans. Souvent le header on s'attend à une sorte d'en-tête (qui contient ou pas de la navigation) donc peut être qu'ici le header serait plutôt la section avec la grosse image, mais c'est vraiment une question de point de vue.

- Dans ta grande image, au lieu de mettre deux H6, peut être mets plutôt un P et règle la taille de font dans ton css si c'est pas assez gros (car H6 on s'attend vraiment à un titre ou un intitulé de section)

- You don't need "div-div" in your H2 (so smart though haha) in the middle section (and in the green as well) because you can create the line with a simple trick : border-bottom ! tadaaa

- Be careful about your indentation : It's nice and clean to put line breaks between sections so your code is readable easily, but the indentation is very important (sometimes a div is too on the left compared to the parent, in the HTML code) it has no incidence on the rendering but for humans reading the code it's harder :)

- In the green footer, you don't need to wrap the H3 in a DIV alone, you can put it in the DIV of all the white text, and then you will have DIV = left text block, IMG = white logo, IMG = right image (not in a DIV but alone as well), which is cleaner. Less HTML = less css blocks = less bugs / faster page !

= = = = = = = = = = = = = = = = = = = = = = = = = = = = = = = = = = = = = = = = = = = = = = = = 

CSS : that's a lot haha but it works, and the result is clean ! Lets try and simplify : 
- * : you don't need to say "border : 0" (it's even weird to say it like that) and by default, there is no border. Nice for the margin / padding though !

- Try and use em / rem for your font sizes (and even padding and margin, that can be cool sometimes) because it's more responsive (1rem "root em" = 16px, and 1em = the font size of the parent element) which means if the screen changes / people zoom, everything will adapt.

- you can't say "justify-content : space around" for the NAV H2 because it's an element, not a box (flex & co only work in containers to organize elements, not on elements themselves)

- #speaker, #quality #devices all have the same properties, why not a .picto class ? And you don't use them since you say ID (#) but they don't have an id, just a alt (description of the image)

.pictograms div = targets the three divs, why width = 100% ? maybe 30% is better non ? (it's the percentage of the parent) and you can't use "align items" in it since you didn't activate flex on them (you also gave them classes that you use later, put one class instead, like .small-box and regroup everything inside it !)

- you repeat "middle part h2" you should regroup them, and don't use vh (vue HEIGHT) on padding because padding is on every side (even horizontal) so use em instead (the horizontal version of vh if vw by the way ! if you want to use it one day). Plus using a screen size unit is not the best choice because you want your padding / margin to be in relation to the size of the text, not the screen !

- your "question" class is weird, call a ".middle-part h2" instead, it makes more sense

- you repeat your class .pictograms later in the css, not good you're going to create conflicts one day haha ! try and regroup everything so you don't repeat yourself (also, flex-direction row is the default option, so you don't need it) and get rid of the previous .pictograms because this one is better (with "baseline" and "space-evenly") !

- in your green footer, you have "padding top", then "padding bottom", then "padding" no no ! use only "padding" instead : value1(top / bottom) value2 (right / left) // [value3 (bottom if different than top) value4 (left if different than right)]



Nice use of flex and classes overall !! You just got confused, and repeated yourself. Try and work on big sections first (display, colors, sizes and all...) then into specific elements. That way you'll se the small details to do and not repeat the big features.

Try to comment your CSS code when you work, it's easier to find out what is where. To comment you do (on a mac) Cmd + Shift + "/" (you can create small parts : NAV / HERO (the big image is called a hero landing) / WHITE SECTION / GREEN SECTION

That way you'll see better what you have :)

Keep practicing there is no secret ! But you have a good eye and your result is very clean, so it's just learning how to make it easier and faster for you !

Tell me if you have any question !
Nina

PR #1835, opened 2020-03-28
PAR WDFT 202003 - Anne Patana - Week1 Day1 Lab ongoing

apatana:
I'm still struggling with the basic concepts such as positioning (absolute, relative, fixed...) and am aware of the lack of responsiveness which I need to start working on...

### DIFF
--- a/index.html
+++ b/index.html
@@ -1,18 +1,116 @@
 <!DOCTYPE html>
 <html lang="en">
-  <head>
-    <meta charset="UTF-8" />
-    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
-    <meta http-equiv="X-UA-Compatible" content="ie=edge" />
-    <title>Spotify Clone</title>
-    <!-- don't forget to link your styles -->
-  </head>
-  <body>
-    Premium Discover Help Download Music for everyone. Spotify is now free on mobile, tablet and computer. Listen to the
-    right music, wherever you are. What’s on Spotify? Millions of Songs There are millions of songs on Spotify HD Music
-    Listen to music as if you were listening live Stream Everywhere Stream music on your smartphone, tablet or computer
-    It’s as yeezy as Kanye West. Search Know what you want to listen to? Just search and hit play. Browse Check out the
-    latest charts, brand new releases and great playlists for right now. Discover Enjoy new music every Monday with your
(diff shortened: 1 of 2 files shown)


TatijanaRajcic:
Hey Anne! Thanks for your work! It looks nice.
- For the responsiveness, do not worry! It was not the main subject of the lab! If you want to practice it's OK but don't worry too much There is a special lab on responsiveness on the student platform called "Slack", in case you want to practice when you have time.
- I think you overcomplicated some things: for example, you did not need your container-id element. The final result should take the whole width of the page, you don't have to artifically recreate a page. So i commented out your container-id and modified width of the navbar to be 100% (of the body), instead of 1000px.
- For your navbar: you do not need that many divs. You could apply several classes to the same div. For example, i applied the class "nav" and the class "fixed" to the same div, so i got rid of one extra div:

```
      <div class="nav fixed">
        <img class="spotify-logo" src="images/spotify-logo.png" alt="spotify-logo">
        <nav class="nav-bar">
          <ul class="nav-bar-ul">
            <li>Premium</li>
            <li>Discover</li>
            <li>Help</li>
            <li>Download</li>
          </ul>
        </nav>
      </div>
```

```
.nav {
    display: flex;
    background-color: white;
    width: 100%;
}
.fixed {
    position: fixed;
}
```

Separating nav and fixed makes sense, if you are reusing those classes somewhere else, otherwhise you could have also merged all those properties, inside one unique class :) 

- I get why you want to imbricate the nav-ul inside a nav tag (to make your HTML correct) but you don't need the nav-bar class, as it does not do anything. The display:flex is to be applied of the ul element, what you basically already do with your nav-bar-ul class. So this one is enough!

- Regarding your landing photo: I saw in your comment that you are struggling with positionning. I can understand, because it is not easy! If you want to position some element, compared to another, you need to : use "position: relative" on the reference element, and "position:absolute" on the element you want to position. You could use that here, to position the landing-text compared to the landing-img, but it would take some time. A simpler solution would be to : 1) remove the img tag from the HTML, 2) use background-image: url(...) on the landing-container, and use the landing image here, 3) use flexbox on the landing-container to center the text vertically and horizontally. If you still want to use positionning, then you could play on the "top" property % to center the text even better.


Nice work Anne :) keep up the good work !

PR #1841, opened 2020-03-30
[PARIS-FT-032020] - Benjamin Dean ongoing

bdean05:
My final spotify commit

### DIFF
--- a/index.html
+++ b/index.html
@@ -5,14 +5,37 @@
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
     <meta http-equiv="X-UA-Compatible" content="ie=edge" />
     <title>Spotify Clone</title>
-    <!-- don't forget to link your styles -->
+    <link rel="stylesheet" href="./styles/style.css">
+    <div class="header"><img src="./images/spotify-logo.png" width="200">
+      <ul class="nav-bar-ul">
+        <li class="nav-bar-item nav-bar-li"><a href='#'>Premium</a></li>
+        <li class="nav-bar-item nav-bar-li"><a href='#'>Discover</a></li>
+        <li class="nav-bar-item nav-bar-li"><a href="#">Help</a></li>
+        <li class="nav-bar-item nav-bar-li"><a href="#">Download</a></li>
+      </ul>
+    </div>
   </head>
   <body>
-    Premium Discover Help Download Music for everyone. Spotify is now free on mobile, tablet and computer. Listen to the
-    right music, wherever you are. What’s on Spotify? Millions of Songs There are millions of songs on Spotify HD Music
-    Listen to music as if you were listening live Stream Everywhere Stream music on your smartphone, tablet or computer
-    It’s as yeezy as Kanye West. Search Know what you want to listen to? Just search and hit play. Browse Check out the
-    latest charts, brand new releases and great playlists for right now. Discover Enjoy new music every Monday with your
-    own personal playlist. Or sit back and enjoy Radio.
+    <div class="content"><h1 class="hightitle">Music for everyone<br /><br /><br /><spam style="font-size: 24px;">Spotify is now free on mobile, tablet and computer.<br />Listen to the
+      right music, wherever you are.</spam></h1></div>
+    <div><h2 class="what">What’s on Spotify?</h2><br /></div>
+    <div class="box">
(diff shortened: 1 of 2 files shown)


TatijanaRajcic:
Hello Benjamin! Thanks for your lab! The page looks nice but there are quite a few things to fix in the code. So here we go:

- You should not have any page content outside the body tag. So please move the div that you have inside the head tag right now in the body. The head tag should only includes global info of your page (its title, the links to the css stylesheets, etc.). 
- Clean your CSS to remove the classes you do not need. It would be easier for your future teammates to know which code is useful :)
- Please avoid inline styling directly in your HTML. It is bad practice as you might loose track of your styling, get confused between the one you write inline and the one you wirte in your css files).
- Please be careful of the identation in your HTML file. Every child element of a parent element should be indented. It will make your code way more readible
- In your navbar: the navlinks are not exactly vertically centered because you added a margin top to the ul ".nav-bar-ul". You could just remove it, and the links would be perfectly vertically aligned.
- "br" tags have a semantic meaning! They should not be used for going to a lne, except if you are in a poem or something like that ^^ . Here, wrapping your text in "p" or in "h1" (if it's a title), would have been enough.
I cleaned your code to this:
```
    <div class="content">
      <h1 class="hightitle">Music for everyone</h1>
      <p>Spotify is now free on mobile, tablet and computer.</p>
      <p>Listen to the right music, wherever you are.</p>
    </div>
```

Then, you'll notice that your elements are in a row now. That's why I changed your class "content" and added "flex-direction: column" to it, so that elements would go one below the other. And then, since I removed the inline styling from your html, I added those properties in the CSS directly: 

```
.content p {
  font-size: 24px;
  line-height:0px;
}
```

- You don't need extra div when you don't make any use of them. For example here:

`    <div><h2 class="what">What’s on Spotify?</h2><br /></div>`

Writing `<h2 class="what">What’s on Spotify?</h2>` would have been enough


- Inside your class "box", you should not create boxes by putting everything inside on div, and spacing them with "br". The easiest way to do that would be, to create a div, with 3 elements inside (image, title, and description). You will achieve the exact same result as your page, but with a working HTML.

```
    <div class="box">
      <div>
        <img src="./images/music-icon.png" width="100">
        <h3 class="bigreen">Millions of Songs</h3>
        <p>There are millions of songs on Spotify</p>
      </div>
      <div>
        <img src="./images/high-quality-icon.png" width="100">
        <h3 class="bigreen">HD Music</h3>
        <p>Listen to music as if you were listening live</p>
      </div>
      <div>
        <img src="./images/devices-icon.png" width="100">
        <h3 class="bigreen">Stream Everywhere</h3>
        <p>Listen to music as if you were listening live</p>
      </div>
    </div>
```

Since you are already using "flex" on your "box" class, I did not have anything to change on the css side. You will notice that the middle box is not perfectly aligned under the title "what's on spotify" . That's because the 3 boxes are not of equal width, so basically, even if you are using "justify-content:space-around" on the parent container, they are spaced around but don't have the exact same width. So simply add a width to those 3 boxes, by adding in your css:

```
.box div {
  width: 33%;
}
```

- Be careful of your indentation in the footer. Moreover, don't use "h4" titles if you don't have "h3" title at some point previously. It's not semantically correct. Moreover, same thing as before: do not use "br" to separate the text. I'll wrap every common text in one div, to make separate groups:

```
    <div class="footer">
      <div>
        <h2 class="under">it's as yeezy as Kanye West</h2>
        <div>
          <h2>Search</h2>
          <p>Know what you want to listen to?</p>
          <p>Just search and hit play.</p>
        </div>
        <div>
          <h2>Search</h2>
          <p>Know what you want to listen to?</p>
          <p>Just search and hit play.</p>
        </div>
        <div>
          <h2>Search</h2>
          <p>Know what you want to listen to?</p>
          <p>Just search and hit play.</p>
        </div>
      </div>
      <div class="app">
        <img src="./images/spotify-icon-white.png" width="100" height="100">
      </div>
      <div class="app">
        <img src="./images/spotify-app.jpg" width="250" height="444">
      </div>
    </div>
```

It is semantically more correct. To be sure that the paragraphs are not too much space, I added in the css: 

```
.footer p {
  line-height: 10px;
}
```

It is a lot of comment but I'm sure you'll get it really fast! Don't hesitate if you have any question !! :+1:

PR #1927, opened 2020-06-17
[Remote-PT-0620] Debora Crescenzo

NubivagantYansa:
Iterations completed. Text alignment green section to be reviewed after Flexbox class.
I imported the Spotify font in the stylesheet, but it doesn't look great in Chrome - Use Safari for best user experience.

### DIFF
--- a/index.html
+++ b/index.html
@@ -1,6 +1,8 @@
 <!DOCTYPE html>
 <html lang="en">
   <head>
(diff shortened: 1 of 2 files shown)


valmgisbert:
You don't have to specify sides if all values are 0. Good work on naming classes. Nice work!

PR #1928, opened 2020-06-17
[REMOTE-WDPT-0602] - Raquel Nascimento

RaquelNascimento96:
- Computer model/size in which it was built: MacBook Pro (13-inch, 2017, Two Thunderbolt 3 ports)


valmgisbert:
Review the navbar links. Try to make your headings hierarchy follow along the skeleton to avoid it being all over the place. Good on naming classes/ids. Nice work!

PR #1931, opened 2020-06-17
[REMOTE-WDPT-0602] - Matthew Howard

mattalanhoward:
Still need to makes some adjustments.  I'm having a bit of trouble with the white space on the sides of the background-image.  Would like to make it responsive.

### DIFF
--- a/index.html
+++ b/index.html
@@ -8,8 +8,8 @@
     <title>Spotify Clone</title>
     <!-- don't forget to link your styles -->
   </head>
+  
   <body>
-
       <nav>
         <img src="./images/spotify-logo.png" alt="Spotify Logo">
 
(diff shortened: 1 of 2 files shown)


mattalanhoward:
A few issues.  I'm unsure of how to fix. 

- White space next to background image in main.

-Extend the bottom border for the yeezy part.  The container width is limited in order to flex those 3 items evenly and center everything.  I took wordwrap off of the phrase so that it would extend past the edge of the container but now I can not get the bottom border (underline) to extend outside of the container.  

-I know I need some @media for the nav and the bottom section but we didn't cover it yet so I didn't put it in!   

How is the overall structure of the html?  What can I do to keep the css DRY?  Thanks!

valmgisbert:
Be careful with syntax errors in naming classes. Inserting imgs in divs helps control their alignment and size. also using percentage sometimes gives risk to inconsistency (especially in the html tag). Good on naming tho. Nice work!
In answering a bit from above: to keep it DRY, you gotta first evaluate your styling to pinpoint exactly which elements are reusable and how to take as much advantage of them with your classes; by naming them as logically as you can, when you connect them to the styling it can be as concise as possible to keep track of it.

PR #1932, opened 2020-06-18
[REMOTE-WDPT-0602] Cristina Matamoro

cris-developer:


valmgisbert:
Good in structure and sizing! Try to find reusable styling inside your skeleton. Try to stay consistent in calling imgs either from CSS or HTML.  Good on naming classes. Good job!

PR #1933, opened 2020-06-18
[REMOTE-WDPT-0620] -Rodrigo Hernández

Rodrigohernandezsan:
Lab completed. I have had some problem in justifying the texts and icons. Thank you


valmgisbert:
Careful with font-sizes and heading hierarchy, also in setting the font-size in the body. Good on naming and using the alt attribute! nice job!

PR #1934, opened 2020-06-18
[REMOTE-WDPT-0620] - Sunitha Mannam

SunithaMannam:
There are some issues in CSS for the footer .and the menu . 
Need to work on it further. 
Not sure of the font-names also

### DIFF
--- a/index.html
+++ b/index.html
@@ -1,18 +1,100 @@
-<!DOCTYPE html>
-<html lang="en">
-  <head>
-    <meta charset="UTF-8" />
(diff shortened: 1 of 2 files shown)


valmgisbert:
Review your url path for the bg image. pPractice your font-sizes hierarchy and `div` enclosing since it can mix certain styles. If all your directions have the same amount you don't need to specify each. Careful with percentages since it can give room to inaccuracies regarding parent elements. Closing imgs in divs can help with size and alignment. Good in naming classes. Nice work!

PR #1935, opened 2020-06-18
[REMOTE-WDPT-0602] Pedro Lamy CSS Spotify Clone

PedroMMLamy:


valmgisbert:
Be careful where you store your links and meta elements; if you don't store it in <head> it won't be recognised. Also in closing sections of styling, otherwise it will be an error. Be consistent in either sourcing from CSS or HTML, depending what it is; mostly to avoid calling it in both files. Try not to use capslock in naming your classes. Good on element structure for your body, and in class selecting for reusable styling. Nice work, tho! Keep mostly in mind syntax and positioning errors to actually see your styling.

PedroMMLamy:
Val thank you for your time and insights. I understand the corrections now.

A terça, 23/06/2020, 19:35, valmgisbert <notifications@github.com> escreveu:

PR #1936, opened 2020-06-18
[REMOTE-WDPT-0602] - ALEJANDRO GONZALEZ

alejanglez:
SPOTIFY CLONE

### DIFF
--- a/index.html
+++ b/index.html
@@ -1,18 +1,84 @@
 <!DOCTYPE html>
 <html lang="en">
   <head>
+    <link rel="stylesheet" href="C:\Users\alejandro.gonzalez\Desktop\code\lab-css-spotify-clone\styles\style.css" />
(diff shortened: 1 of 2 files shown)


valmgisbert:
Try to select a class for the divs that keep your imgs so you don't lose track of them, or stay consistent in adding classes to your imgs. Avoid using <br> so much, especially if it's used for stylistic purposes like here. Careful with the spacing in your HTML skeleton. If you are dividing your skeleton in sections, try to stay consistent and just leave out other main parts of the body. Remember that <divs> are mostly for stylistic purposes, not exactly recommended for actually grouping your code if you want to structure it within the skeleton. Avoid starting attributes with capslock.  Good on calling classes. For what I asked in this exercise, it's okay but if you want to practice styling, work on your skeleton.

PR #1937, opened 2020-06-18
[REMOTE-WDPT-0602]-Nuno Pacheco

nuno-pacheco:
I wish I have more time to make it perfect, but I think I'm getting the basics ;)


valmgisbert:
What a clean, well-selected skeleton! Good on naming and referencing your elements. Try to find certain styles that repeat themselves in your css and group them in a way that helps you maintain reusable code; one element can have more than one class! use that to your advantage to be as organised and concise as possible. You're really grasping the basics! More practice and you'll be even better! Great job!

PR #1939, opened 2020-06-18
[REMOTE WD PT JUNE 2020] - Andre Bacalhau

AndrewMastercode:


valmgisbert:
If you're maintaining your styling in your main.css, avoid hardcoding styles in your HTML. Work a bit on your spacing for easier-to-read code. If you have a section that has small clusters of info, you can insert them in divs for easier way of handling into styling; also, following this, avoid inserting sections within sections since we use them to divide groups of elements that might fit one theme. In your main.css, avoid leaving elements blank/without values. Good on naming and referencing. Nice job!

PR #1940, opened 2020-06-18
[REMOTE-WDPT-0602] - Juan Jose Carreño

JJexcelsior:


valmgisbert:
Try to avoid using sections within sections, since these are used more for the structure in the skeleton to keep one group of elements that will be following a single theme along your HTML; if you want to use this in a style-assisting way, use divs! they are perfectly ready to be implemented just to group stuff for styling purposes. Good in naming and referencing! And it's responsive! Very nice! Great work!

PR #1958, opened 2020-07-09
[AMS - PT - June20] - Lisette & Kenouly

Kenouly:


styles/style.css
```diff
@@ -6,3 +6,110 @@ Green: #00B172
...
+    text-align: center;
+}
+
+.intro h1 {
+    padding-top: 350px;
+}
+
+.services h3 {
+    text-align: center;
+    text-decoration: underline #00B172;
```
guido-an, line 54: In this way the green divider takes all the screen width and the that's not our goal here.
You could wrap the h3 inside a span ` <h3> <span>What’s on Spotify?</span></h3>` and then use border-bottom: 
`.services h3 span {`
  `  border-bottom: 4px solid #00b172;`
`}`

PR #1960, opened 2020-07-10
[AMS-WDPT-June 20] Yuka  

yukajyeh:
90% done in class pair-coding. 
For the 'Yeezy' part I think there are better ways to position items, would like to receive feedback especially for this part :)


index.html
```diff
@@ -5,14 +5,79 @@
...
+     <ul class="nav-list">
+       <li>Premium</li>
+       <li>Discover</li>
+       <li>Help</li>
+       <li>Download</li>
+     </ul>
+    </nav>  
+  </header>
+
+  <section class = 'section-1'>
```
guido-an, line 24: Remove extra spaces, here it should be `<section class='section-1'>`

guido-an: Same in the following sections

guido-an:
Really good Yuka!

PR #1962, opened 2020-07-10
[WDPTAMS062020] Anna Dorenskaia

AnnaDore:
I was working in a really good collaboration with Marcella. Together we did 95% (except footer). Unfortunately, css is quite difficult for me, during the lesson for me it was really difficult to suggest any good solutions. I decided it would be unfair to put the solution without an understanding of it and I decided to redo everything for practice. 
I'll do my best to fix my current pure level of css.

### DIFF
--- a/index.html
+++ b/index.html
@@ -10,8 +10,8 @@
   </head>
   <body>
     <header class="header">
-      <div>
-        <img src="/images/spotify-logo.png" alt="logo spotify">
+      <div class="nav">
+        <img class="header-icon" src="/images/spotify-logo.png" alt="logo spotify">
         <div class="header-list">
           <ul>
             <li>Premium</li>
@@ -20,6 +20,8 @@
             <li>Download</li>
           </ul>
         </div>
+      </div>
+      <div class="header-text">
         <h1>Music for everyone.</h1>
         <p>Spotify is now free on mobile, tablet and computer. Listen to the
           right music, wherever you are.</p>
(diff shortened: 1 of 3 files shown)


index.html
```diff
@@ -5,14 +5,62 @@
...
   <body>
-    Premium Discover Help Download Music for everyone. Spotify is now free on mobile, tablet and computer. Listen to the
-    right music, wherever you are. What’s on Spotify? Millions of Songs There are millions of songs on Spotify HD Music
-    Listen to music as if you were listening live Stream Everywhere Stream music on your smartphone, tablet or computer
-    It’s as yeezy as Kanye West. Search Know what you want to listen to? Just search and hit play. Browse Check out the
-    latest charts, brand new releases and great playlists for right now. Discover Enjoy new music every Monday with your
-    own personal playlist. Or sit back and enjoy Radio.
+    
+    <header>
+      <section class='navbar'>
```
guido-an, line 13: You could use <nav> here instead of <section>

styles/style.css
```diff
@@ -6,3 +6,128 @@ Green: #00B172
...
+}
+
+.header-background {
+    background-image: url(/images/landing.jpg);
+    background-size: 100%;
+    background-repeat: no-repeat; 
+    height: 700px;
+    background-position: 50% 50%;
+    position: relative;
+    color: white;
```
guido-an, line 49: If you use _text-align: center_ here there is no need to use _position: absolute_ later.
For example: `.header-background {`
    `background-image: url('/images/landing.jpg');`
   ` background-size: 100%;`
    `background-repeat: no-repeat;`
   ` height: 700px;`
  `  position: relative;`
    `color: white;`
  `  text-align: center`
}

`.header-background>p {`
  `  font-size: 2em;`
   ` text-align: center;`
`}`

`.header-background h1 {`
    `padding-top: 200px;`
   ` font-size: 4em;`
`}`

PR #2016, opened 2020-08-25
Made the clone of Spotify

JorellLangmar:
Hey !

You'll find there the index.html and the index.css used to replicate the spotify webpage.

Cheers

### DIFF
--- a/index.html
+++ b/index.html
@@ -4,15 +4,70 @@
     <meta charset="UTF-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
     <meta http-equiv="X-UA-Compatible" content="ie=edge" />
+    <link rel="stylesheet" href="./styles/style.css">
     <title>Spotify Clone</title>
     <!-- don't forget to link your styles -->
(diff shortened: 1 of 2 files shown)


N10T:
Hey Jorell,
I forget to say it. The title has to be like this : [PAR-WDFT-248] YOUR NAME

In general, the page looks good but...
- under the title, you write it in 3 like. It was in 2 lines.
- in the green section, the Spotify logo is not centred.

HTML:
- In your nav, you don't need to use ul and li, it's an extra work

CSS:
Initialisation : 
* {
  margin: 0;
  padding:0;
  box-sizing: border-box;
}

- line 53: prefer to write the path as a string => background-image: url("../images/landing.jpg");
- .right => .infos and .left => .screenshot
-#yeezy div line 136: I think you target more than what you what, prefer #yeezy>div
- easier to center your spotify logo, use it as a background:  

#yeezy {
  background-repeat: no-repeat;
  background-size: 120px;
  background-position: 50%;
}

👍 good job

PR #2017, opened 2020-08-25
[PAR-WDFT-248] Luca Marongiu

AyloSrd:
I don't know if the font-family and the font-size are correct, but I think the result looks enough like the original.
Best regards,
LM

### DIFF
--- a/debug.log
+++ b/debug.log
@@ -1,0 +1,2 @@
+[0825/155744.274:ERROR:directory_reader_win.cc(43)] FindFirstFile: Le chemin d’accès spécifié est introuvable. (0x3)
+[0825/165640.899:ERROR:directory_reader_win.cc(43)] FindFirstFile: Le chemin d’accès spécifié est introuvable. (0x3)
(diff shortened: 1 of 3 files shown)


N10T:
Hey it looks the same 🥇 

CSS:
- write your names in kebab case "greenVerbsSection" => "green-verbs-section"
- your nav contain li but no a tag, your nav is to navigate, you need more a tags than li tags
- flexContainer good idea, but you use it only in #navbarList in this case don't create a class and a id when it's used for the only same thing
- "--spotify-green" 💯 good practice to use variables

HTML:
- Your green yeezy part in not nicely ordered: the best is to draw it before code it.
<footer id="footer_main" class="row">
      <div class="infos">
        <h4 class="title underlined white">It’s as yeezy as Kanye West.</h4>
        <h5 class="title">Search</h5>
        <p class="text">
          Know what you want to listen to? Just search and hit play.
        </p>
        <h5 class="title">Browse</h5>
        <p class="text">
          Check out the latest charts, brand new releases and great playlists
          for right now.
        </p>
        <h5 class="title">Discover</h5>
        <p class="text">
          Enjoy new music every Monday with your own personal playlist. Or sit
          back and enjoy Radio.
        </p>
      </div>

      <div class="wrap">
        cool pic
      </div>
    </footer>

My proposal, in this case you don't need to fix the position of something at the end.

PR #2018, opened 2020-08-25
[PARIS-FT-August2020]Laura-Caillaux

Lcdevi:


Fanny-Sbn:
Hey Laura, thank you so much for your lab, it’s great ! Just a few comments 😊

The big differences between the original site and your version are: 
The nav bar is a bit bigger in the original site (the logo either)
In the what’s on part, your subdivisions should take more space. 

HTML:

•	In your nav, you don't need to use ul and li, you should have used a <nav> tag and <a> tags for the links, basically (just an example of how it can look) :

<nav id=”nav”>
        <a class="link">Premium</a>
        <a class="link">Discover</a>
        <a class="link">Help</a>
        <a class="link">Download</a>
 </nav>

•	You use a lot of <br> tags in your html code, for instance in the last part you could have played with font-size and width in your paragraphs to avoid them. 

CSS: 

I saw that you put direct styling in your HTML file, it’s okay but if you have created a css file for your style it’s better to be consistent and assign a class or id to the element you want to target and put its style in your css file. 

•	The properties display:flex and flex-direction:column in your body tag are useless
•	Use kebab case "boxsongs" => "box-songs" for your ids and classes
For what you call the Kanye part, a few things:
•	Use flexbox and a justify-content: space-between to place on the left side the text part and on the right side the image without using position absolute. 
•	To center your spotify logo, use it as a background:
#kanye {
background-repeat: no-repeat;
background-size: 120px;
background-position: 50%;
}
Keep on going girl ! 💃

PR #2019, opened 2020-08-25
[PAR-FT-082020]Frederico Ratsimandresy

Ratsimandresy:


N10T:
Hello Rico,

Behaviour:
Looks good 🥇 but...
You can have more space around your Spotify logo and after download
Be careful with text mistakes: "It'sa yeezy as Kanye west."
Your green is not the same, colour pick the good one maybe on the real website of spotify. "rgb(29, 185, 84)"

HTML:
In your nav it's harder to code a ul and li, just for a tags

CSS:
- use kebab case "imgsize" => "img-size"
- use explicit names "ptxt","txt","plogo","imgsize"
- body don't need a height
- margin: 15px 0px 0px 0px; => margin-top:15px

Ratsimandresy:
Hi Andy!

Ok thank you 🙏🏼 i will make those correction.

PR #2020, opened 2020-08-25
[PAR-WDFT-248] NICOLAS ENAULT

nenault:
First attempt at grid! 

The page is quite responsive (except for the menu that jumps below a certain size, for laziness reasons :))

### DIFF
--- a/index.html
+++ b/index.html
@@ -4,15 +4,80 @@
     <meta charset="UTF-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
     <meta http-equiv="X-UA-Compatible" content="ie=edge" />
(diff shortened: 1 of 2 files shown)


N10T:
Hello Nico,

Behaviour:
Looks good 🥇 but...
- a bit too much height for your img
- nice grid, your page is responsive !!! 💯 you just have to turn in a column your nav and header

HTML:
- good idea to change the title 👍 

CSS:
- use kebab case for naming 👎 
- multiple classes 👍
- grid-template-columns: 1fr repeat(3, minmax(auto, 25em)) 1fr; 👍 
- use id tags to reference unique sections like "whatsOn", "largeImg"

very good job!

PR #2022, opened 2020-08-25
[PAR-FT-240820] Laura Mantal-d'Aboville

Lauradabo:
done and dusted!


Fanny-Sbn:
Hey Laura, thank you so much for your lab, it looks great! Just a few comments 😊

General observations: 
•	The nav links could have been a bit bigger
•	The text in some subdivisions in the what’s on part don’t break at the right place, you could have used the <br> tag in that case. 
•	Same that the previous observation for texts in the last part, in that case you could have played with font-size and width of your paragrahs

HTML:
•	In your nav, you don't need to use ul and li, you could have used only the <nav> tag and the<a> tags for the links
•	Don’t forget to autoformat your code, it will make your life easier !

CSS: 

So nice you added comments in your code 😊

I’ve seen you’re using tags to target elements, it’s okay for now, but take the habit to attribute an id or a class to them, especially for those which have children, for instance for the <header> tag
For what you call the #bottom-part, a few things:
•	Use flexbox and a justify-content: space-between to place on the left side the text part and on the right side the image
•	To center your spotify logo, use it as a background and remove the  from the html code and the position: absolute in css. 
.bottom-part {
background-repeat: no-repeat;
background-size: 120px;
background-position: 50%;
}
Good job Laura !

PR #2024, opened 2020-08-25
[PAR-WDFT-248] Vincent PROUM

VProum:


N10T:
Hello Vincent,

The behaviour:
Your page is not too far from the original 👍  but...
- Nav text is too small
- Not the same font => inspect the pdf ;)
- "What’s on Spotify?" part, the title and element are not centred
- For the green section, blocks are not vertical centred

HTML:

CSS:
- really clean CSS 🥇 
- section 2, to equally spaced every child elements, they all need the same width => width or "flex: 1"
- kebab case and naming, but you can be more explicit than "section-three" => "green-footer"👍 
- to center vertically your flex element "align-items: center" for the #section-three

PR #2025, opened 2020-08-25
[PAR-WDFT-248] Ailie McCorkindale

RaspberryChaos:


N10T:
Hello Ailie,

The behaviour:
- your page looks like the original 👍 but...
- "What’s on Spotify?" the original height is taller and your children have too much width
- the original green footer :
- has a margin
- the logo is not centred
- the left part looks too large (because of the text)
- the z-index of the header has to be 1

HTML:

CSS:
- good font choice 👍 
- clean CSS 💯 
- kebab case 👍 
- Spotify-icon use it as a background of the #kanye part

PR #2026, opened 2020-08-25
[PAR-WDFT-248] LISA PONCET

lisapnct:
pain points: 

- position the image in the header background and adjust its size
- the last green part: i struggled defining the right layout and got lost in my divs i think :(
- position:absolute of the white spotify icon

### DIFF
--- a/index.html
+++ b/index.html
@@ -62,6 +62,7 @@
     <!-- FEATURES SECTION -->
     <section id="features">
       <h2>It’s as yeezy as Kanye West.</h2>
+      <img id="logo-spotify-white" src="./images/spotify-icon-white.png" alt="spotify-icon-white">
 
       <div class="features-div">
 
@@ -83,10 +84,6 @@
               with your own personal playlist.
               Or sit back and enjoy Radio</p>
           </article>
-        </div>
-
-        <div id="img-item-logo">
-          <img src="./images/spotify-icon-white.png" alt="spotify-icon-white">
         </div>
 
         <div id="img-item-app">
--- a/index.html
+++ b/index.html
@@ -63,30 +63,37 @@
     <section id="features">
       <h2>It’s as yeezy as Kanye West.</h2>
 
-      <div class="vertical-flex-container">
-        <article class="flex-column-item">
-          <h3>Search</h3>
-          <p>Know what you want to listen to?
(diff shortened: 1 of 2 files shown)


Fanny-Sbn:
Hey Lisa, thank you so much for submitting your lab ! Just a few comments 😊

General observations: 

•	The nav bar should take the all width of the page, you have a small space on the right
•	The nav links and the spotify logo could have been a bit bigger
•	The what’s on spotify part is not big enough, same for the margin and the titles
•	The text in some subdivisions in the what’s on spotify part don’t break at the right place, you could have used the <br> tag in that case. 
•	In the green part, same that the previous observation for texts, in that case you could have played with font-size and width of your paragraphs. Titles are not big enough.
•	Your spotify icon is not centered, but you’ve already mentioned it 😉

HTML:
•	In your nav, you don't need to use ul and li, you could have used only the <nav> tag and the<a> tags for the links, it could look like something like that : 

<nav id="navbar">
        	<a class=" navigation-btn">Premium</a>
        	<a class=" navigation-btn ">Discover</a>
        	<a class=" navigation-btn ">Help</a>
       	 <a class=" navigation-btn ">Download</a>
</nav>

•	About section is great 
•	For the last section (green part one) : you need to remove the <div class="features-div", it’s useless and you can move your <h2> inside of the <div class="vertical-flex-container">

CSS: 
•	No need to set a min-height for the body here. 
•	In your nav bar, you can add a width so it can take the whole page width 
.fixed-navbar{
width: 100vw;
	}
•	For positioning the image and centering the text in it, have a look to this link: https://www.w3schools.com/howto/howto_css_hero_image.asp
•	Add a padding-bottom and padding-top to your <div class="horizontal-flex-container"> and try to resize the elements inside your articles so that it can look closer to the original spotify website.

•	Use flexbox and a justify-content: space-between to place on the left side the features-div part and on the right side the image
•	To center your spotify logo, use it as a background and remove the  from the html code and the position: absolute from css, it will make your life easier
. features-div {
background-repeat: no-repeat;
background-size: 120px;
background-position: 50%;
}
•	I notice that you tried to directly line wrap in your html code for the paragraphs, note that it will not work in your browser using this way. Try to resize all the elements (titles and paragraphs with font-size and div with width) in your features-div. 

PR #2027, opened 2020-08-25
[PAR-WDFT-248] FABIEN MARCHAND

fabienMarchand:
First commit

### DIFF
--- a/index.html
+++ b/index.html
@@ -4,15 +4,96 @@
     <meta charset="UTF-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
     <meta http-equiv="X-UA-Compatible" content="ie=edge" />
+    <link rel="stylesheet" href="/styles/style.css">
     <title>Spotify Clone</title>
     <!-- don't forget to link your styles -->
   </head>
+
   <body>
-    Premium Discover Help Download Music for everyone. Spotify is now free on mobile, tablet and computer. Listen to the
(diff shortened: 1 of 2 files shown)


N10T:
Hey Fabien,

Your page looks like the original but you have not finished it, or not pushed it before pull request 👎, **you have to finish it**
You can try to rush it but don't pull request with no reason unfinished work
 
the behaviour:
- the text landing text is not centred
- use the same green everywhere
- **You have not finished the green footer**
- the green is not the same everywhere

HTML:

CSS:
- don't forget to use kebab case for naming in CSS "welcomeContainer"
- z-index: 200; no need to go so high  :)
- use this Spotify icon as a background for the green footer
- sectionFeatures don't need a position relative with good flexbox

fabienMarchand:
Hey Andy,

I update mySpotify,

I just do a comma and push is it enough for you to check my correction ?

Tell me if you think I have to change things before saying to that is good enough

Thanks in advance

Fanny-Sbn:
Hello Fabien, thanks for the modifications. 

It looks way better ! Just a few comments.

HTML : 

-You have an extra div in your header, the direct one. You can simply move its properties in the header ;)
-You don't need to use ul and li tags, the nav tag and a tags are enough ! (don't forget to modify the CSS in consequence)

CSS : 

Good ! 

-Use the border-bottom property for the yeezy title. 

Nice job !

PR #2028, opened 2020-08-26
[PAR-WD-248] BOUZIANE BEY

Bouziane95:


N10T:
Hello Bouziane,

**The behaviour:**
Your page is not too far from the original 👍 but...
- The text of all the page is too small
- "What's on Spotify ?" is not underlined
- "It's as yeezy as Kanye West" part is not centred vertically, use flexbox properties

**HTML**
- <br> use it when you need it, maybe not after the h3
- you have 3 sections and the second one have id "first-section"...


**CSS**
- kebab case 👍 
- "green-section" explicit name 👍 "#column-div", "title-green-section-underline" not explicit and super long names 👎 
- ".title-green-section-underline" can be #first-section h3:first-child
- this part style nothing:
  #spotify-green-section{
      display: block;
      margin-left: auto;
      margin-right: auto;
}

PR #2030, opened 2020-08-26
WDFT 248  / Philippe Buteau / Lab Spotify Clone

s3chs:
I couldn't get the border-bottom to work on the last section :/

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,15 @@
+{
+    // Use IntelliSense to learn about possible attributes.
+    // Hover to view descriptions of existing attributes.
+    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
+    "version": "0.2.0",
+    "configurations": [
+        {
+            "type": "pwa-chrome",
+            "request": "launch",
+            "name": "Launch Chrome against localhost",
+            "url": "http://localhost:8080",
+            "webRoot": "${workspaceFolder}"
+        }
+    ]
+}
--- a/index.html
+++ b/index.html
@@ -70,32 +70,34 @@
 
 <section id="kanyeWest">
 
-<img class="logoCenter" src="./images/spotify-icon-white.png" alt="spotify-icon-white">
+<div class="blocks">
 
-  <div class="howTo">
     <h2>it's as yeezy as Kanye West<h2>
-  </div>
 
-  <div class="howTo">
     <h3>Search</h3>
     <p>Know what you want to listen to ?</p>
-  </div>
 
-  <div class="howTo">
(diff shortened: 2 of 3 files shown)


Fanny-Sbn:
Hey Philippe, thank you so much for submitting your lab.
 
General observations:

• The nav links and the spotify logo could have been a bit bigger and your logo is deformed (be careful with the dimension of an image)
• The what’s on spotify part is not big enough, same for the margin and the title, you have no space between the big image and the title.
• The text in some subdivisions in the what’s on spotify part don’t break at the right place, you could have used the br tag in that case.
• In the green part, same that the previous observation for texts, in that case you could have played with font-size and width of your paragraphs. Titles are not big enough.
• Still in the green part, the big title is not underlined and you don’t even put the capital letter at the beginning.
• You don’t have any margin-bottom for the green part, on the original version there is one. 

HTML:
You’re giving a class name to your header but you never use it so what’s the point ? You should have used it instead of using directly the tag header in your CSS
• In your nav, you don't need to use ul and li, you could have used only the nav tag and the a tags for the links
• For the second part, the section you called description, you need to move the h2 with the id title in it and wrap all the other div which have the class small-container with a div.
• In your divs small-container your using a h4 for tag and it is semantically incorrect, it’s not a title so you should use a p tag instead

• For the last section (green part one) great structure 😊
• Last point: you had a footer tag with a class nothing-to-add-here, if you have nothing to add and it doesn’t do anything, just remove it. Your code will be cleaner 

CSS:
• Nice for the initialisation
• use kebab case for your class name and don’t begin your class names with a capital letter
•The landing image should be the background of your div textOver and not of your div landing page
• In your header and nav ul, you don’t need to specify the flex-direction: row, with a display: flex, the default value is row, so it’s an extra line for nothing.
• For your logo: set either the width or the height, not both at the same time, it will keep the right image dimensions.
• You repeat twice the same CSS rule with nav… Be careful !
• For positioning the image and centering the text in it, have a look to this link: https://www.w3schools.com/howto/howto_css_hero_image.asp and modify your div textOver in consequence
• Add a padding-bottom and padding-top to your section description and try to resize the elements inside your divs small-container so that it can look closer to the original spotify website.
• To underline the title h2, have a look to the CSS property border-bottom
• Use flexbox to display correctly the elements in the div blocks
• Try to resize all the elements (titles and paragraphs with font-size and div with width) in your blocks div.
• To center your spotify logo, use it as a background and remove the img  from the html code, it will make your life easier
#kanyeWest {
background-repeat: no-repeat;
background-size: 120px;
background-position: 50%;
}
You need to work a bit on the lab to make it better ! 

PR #2031, opened 2020-08-27
[PAR-WDFT-248] Olivo Raphaëlle

RaphOL:
lab 1 spotify done


Fanny-Sbn:
Hello Raphaëlle, thanks for submitting your lab ! A few comments 😄 

General observations : 

- The landing image should be placed after the nav, currently it's under it without scrolling
- The what's on sportify part in not big enough, you should have add some padding on top and bottom
- The text in some subdivisions in the what’s on spotify part don’t break at the right place, you could have used the br tag in that case.
- In the green part, same that the previous observation for texts, in that case you could have played with font-size and width of your paragraphs. Titles are not big enough.
- You don’t have any margins for the green part, on the original version there are some.

HTML:

- Yould have use a header tag to wrap the img and nav and use a nav tag and a tags instead of ul and li for the links. 
-The rest is ok

CSS:
-Nice for the initialisation
-Add a higher padding-bottom and padding-top to your div bloc2 and try to resize the elements inside  so that it can look closer to the original spotify website.

-Try to resize all the elements (titles and paragraphs with font-size and div with width) in your blocks div.
-Add margins to your last bloc
-To center your spotify logo, use it as a background and remove the img from the html code, it will make your life easier
#bloc3 {
background-repeat: no-repeat;
background-size: 120px;
background-position: 50%;
}

PR #2032, opened 2020-08-29
[PAR - WDFT - 248] Timothée NICOLE

timothee-nicole:
Done! but still strugling with background image size

### DIFF
--- a/index.html
+++ b/index.html
@@ -5,14 +5,78 @@
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
     <meta http-equiv="X-UA-Compatible" content="ie=edge" />
     <title>Spotify Clone</title>
-    <!-- don't forget to link your styles -->
+    <link rel="stylesheet" href="/styles/style.css">
   </head>
   <body>
-    Premium Discover Help Download Music for everyone. Spotify is now free on mobile, tablet and computer. Listen to the
-    right music, wherever you are. What’s on Spotify? Millions of Songs There are millions of songs on Spotify HD Music
-    Listen to music as if you were listening live Stream Everywhere Stream music on your smartphone, tablet or computer
-    It’s as yeezy as Kanye West. Search Know what you want to listen to? Just search and hit play. Browse Check out the
-    latest charts, brand new releases and great playlists for right now. Discover Enjoy new music every Monday with your
-    own personal playlist. Or sit back and enjoy Radio.
-  </body>
+    <!-- Iteration 1 -->
+<nav class="navBar">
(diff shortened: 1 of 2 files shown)


Fanny-Sbn:
Hey Timothée, thank you so much for submitting your lab.

**General observations**
-
• The nav links could have been a bit bigger and your logo is too big (be careful with proportions)
•As you said, your landing image doesn’t take the all width and the title is not big enough either the baseline which is supposed to be on two different lines.
• The what’s on spotify part is not big enough.
• The text in subdivisions in the what’s on spotify part are not big enough and don’t break at the right place, you could have used the br tag in that case 
• In the green part, same that the previous observation for texts, in that case you could have played with font-size and width of your paragraphs. Titles are not big enough.
• Still in the green part, the big title is not fully underlined and isn’t at the right place.
• You don’t have the good margin-bottom for the green part, on the original version there is one bigger.
• Your white spotify icon isn’t centered.

**HTML**
-
•Instead of using a nav tag for the first part use a header tag and use a nav bar for the links with a tags inside instead of using ul and li. 
•Your HTML code is not valid, you’re using a section with a class afternav that is not closed. You can remove it, it’s totally useless.
•Give explicit name to your class and never begin your id or calass name with a capital letter -> iteration2 doesn’t mean anything
• For the last section (green part one), put your h2 outside the div class “feature”, it will be easier to manage. 
•Wrap your spotify app image in a div
•Be careful you have two closing div tags in your HTML code that’s strictly doing nothing and have no opening tags…. Line 77-78

**CSS**
-
• Initialize your CSS file, you won’t have unexpected margin and padding given automatically by the browser you’re using. 
*{
    padding:0;
    margin:0;
    box-sizing:border-box;
}

•Your using HTML tags in your CSS, try to use class or ids, you have several conflicts (h2 for instance)

•You haven’t use all the needed properties to set correctly your background image, you need width and background-size: cover;
• For positioning the image have a look to this link: https://www.w3schools.com/howto/howto_css_hero_image.asp and use flexbox to center the the tile and the baseline
• For your h2 give it an id and don’t use margin to center your text.. 
• Add a padding-bottom and padding-top to your section iteration3 and try to resize the elements inside your divs so that it can look closer to the original spotify website.

• Don’t use a width for your h2 
• Try to resize all the elements (titles and paragraphs with font-size and div with width) in your feature divs.
• To center your spotify logo, use it as a background and remove the img from the html code, it will make your life easier
.kanye {
background-repeat: no-repeat;
background-size: 120px;
background-position: 50%;
}

You need to work again on this lab to make it better!
Wax on, wax off! Keep on going! 💪

PR #2033, opened 2020-08-29
[PAR-WDFT-248] M.FUXA

KiwiMantelli:
Hey,

My spotify clone css and html

### DIFF
--- a/index.html
+++ b/index.html
@@ -1,18 +1,130 @@
 <!DOCTYPE html>
 <html lang="en">
-  <head>
-    <meta charset="UTF-8" />
-    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
-    <meta http-equiv="X-UA-Compatible" content="ie=edge" />
-    <title>Spotify Clone</title>
-    <!-- don't forget to link your styles -->
-  </head>
-  <body>
-    Premium Discover Help Download Music for everyone. Spotify is now free on mobile, tablet and computer. Listen to the
-    right music, wherever you are. What’s on Spotify? Millions of Songs There are millions of songs on Spotify HD Music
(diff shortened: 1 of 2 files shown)


Fanny-Sbn:
Hey Marina, thank you for submitting your lab.

**General observations**
-
• The nav links are supposed to be on the right side of the screem but they’re just next to spotify logo. 
• You have this king of margin all around your page, you need to remove it.
•On the landing image the title is too big...
• The what’s on spotify part is great ! 
• The text in subdivisions in the what’s on spotify part are not big enough and don’t break at the right place, you could have used the br tag in that case 
• In the green part, the texts in your paragraph don’t break at the right place, you could have played with font-size and width of your paragraphs and the title is not aligned to your smaller titles and paragraphs

**HTML**
-
•Wrapping your nav tag with your nav-bar div is useless, that’s why your space-between doesn’t work.  
• For the last section (green part one), you have some extra divs, try to simplify your HTML structure: you need a div to wrap the text part and the image which represents each one a div. Inside the text part you need other tags, one for the title and one for each feature with the article tags as you’ve already done it. 

**CSS**
-
• Initialize your CSS file, you won’t have unexpected margin and padding given automatically by the browser you’re using. 
*{
    padding:0;
    margin:0;
    box-sizing:border-box;
}

• • I notice that you tried to directly line wrap in your html code for the paragraphs, note that it will not work in your browser using this way. Try to resize all the elements (titles and paragraphs with font-size and div with width) in your article tags.
• To center your spotify logo, use it as a background and remove the img from the html code, it will make your life easier
#kanye {
background-repeat: no-repeat;
background-size: 120px;
background-position: 50%;
}

You need to work again on this lab to make it better!
Wax on, wax off! Keep on going! 💪

PR #2034, opened 2020-08-29
[PAR-WDFT-248] MAXIMILIAN GRUBER

MaxJGruber:
Hello,

Please find attached my pull request.

Best regards,

Max

### DIFF
--- a/index.html
+++ b/index.html
@@ -68,28 +68,34 @@
     </section>
     <section class="spotify-description">
       <div id="logo-in-the-bg">
-        <div class="text-about-spotify">
-          <h3 id="title-section4">It’s as yeezy as Kanye West.</h3>
-          <h4 class="what-you-can-do">Search</h4>
-          <p class="what-you-can-do-p">
-            Know what you want to listen to? Just search and hit play.
-          </p>
-          <h4 class="what-you-can-do">Browse</h4>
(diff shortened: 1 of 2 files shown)


Fanny-Sbn:
Hello Maximilian, thanks for submitting your lab.

**General observations**
-
• Your spotify logo and your nav links are not big enough, try to take a bit more space for your nav bar .
•On the landing image the title and the baseline are not big enough either. Plus, your landing image is already behind your nav bar before scrolling, it shouldn’t. 
• For the what’s on spotify part, your subdivisions are too small and the text in it does’t break at the right place, you could have used the br tag in that case 
• In the green part, same than previous observation, the texts in your paragraphs don’t break at the right place, you could have played with font-size and width of your paragraphs in that case. 

**HTML**
•No need to use ul and li for the nav, you could have used the nav tag and a tags inside.
•You have an extra-div for the landing image, you could remove the div main-ad and move its CSS properties in the section #header.
• For the last section (green part one) same comment than before, you could have removed your #logo-in-the-background div and set its properties to the .spotify-description section.


**CSS**
-

• Nice for the initialisation
•Nothing specific to say, except that you need to resize a bit your paragraphs and the texts generally to make it look closer to the original website

Good job! 💪

PR #2035, opened 2020-08-30
[PAR-WDFT-248] SALIM AMEUR 

ameursalim:
with a bit of delay, but i wanted to review all css course before finishing it.
the web-site looks good , but i still struggle to make it responsive.

### DIFF
--- a/index.html
+++ b/index.html
@@ -5,14 +5,105 @@
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
     <meta http-equiv="X-UA-Compatible" content="ie=edge" />
     <title>Spotify Clone</title>
+    <link rel="stylesheet" href="/styles/style.css">
     <!-- don't forget to link your styles -->
   </head>
   <body>
-    Premium Discover Help Download Music for everyone. Spotify is now free on mobile, tablet and computer. Listen to the
-    right music, wherever you are. What’s on Spotify? Millions of Songs There are millions of songs on Spotify HD Music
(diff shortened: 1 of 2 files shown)


Fanny-Sbn:
Hello Salim, thanks for your lab. 

**General observations**
-
You're website is looking good ! But 
-Your logo and the links are not big enough
-The what's on spotify title is not big enough either
-The text in the subdivisions in the what’s on spotify part don’t break at the right place, you could have used the br tag in that case.
-In the green part, same that the previous observation for texts, in that case you could have played with font-size and width of your paragraphs. Titles are not big enough.

**HTML**
- Nice HTML structure 
-In your nav, you don't need to use ul and li, you could have used only the nav tag and the a tags for the links.

**CSS**
-
-Nice for the initialisation .
-Don't use float to place your elements, it's way easier with flexbox ;)

- Use flexbox and a justify-content: space-between to place on the left side the footer-left and on the right side the image
- To center your spotify logo, use it as a background and remove the  from the html code, it will make your life easier
footer {
background-repeat: no-repeat;
background-size: 120px;
background-position: 50%;
}
- I notice that you tried to directly line wrap in your html code for the paragraphs, note that it will not work in your browser using this way. Try to resize all the elements (titles and paragraphs with font-size and div with width) in your footure-box.

-To make your website responsive you need to use media queries, have a look to the course on your student platform. (or to use grid)

Keep on going ! 💪

PR #2102, opened 2020-10-13
[FTAMS102020] Sophie Tichelaar

Sophietic:
done with html structure

### DIFF
--- a/index.html
+++ b/index.html
@@ -6,14 +6,18 @@
     <meta http-equiv="X-UA-Compatible" content="ie=edge" />
     <title>Spotify Clone</title>
     <!-- don't forget to link your styles -->
+    <link rel="stylesheet" type="text/css" href="styles/style.css" />
   </head>
   <body>
+
+
 <!-- Navigatie bar met plaatje  -->    
 <section>
   <!-- Blauwe spotify bar  -->    
 <div>
-  <img src=images/spotify-logo.png alt="Spotify Logo"> 
-  <nav><ul>
+  <nav class="nav-bar">
+    <img class ="greenlogo" src=images/spotify-logo.png alt="Spotify Logo">
+    <ul class="nav-bar-ul">
     <li><a href="#premium">Premium</a></li>
     <li><a href="#discover">Discover</a></li>
     <li><a href="#help">Help</a></li>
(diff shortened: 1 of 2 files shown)


styles/style.css
```diff
@@ -6,3 +6,178 @@ Green: #00B172
...
+
+h3{
+    color: #00B172;
+    font-weight: bold;
+}
+
+h5{
+    font-size: 2em;
+}
+
```
LloydC, line 87: It looks like you defined CSS for header 3 times, you should remove two of them and have the relevant properties in the one left

`header {
     background-size: cover;
     background-position: bottom;      
    padding-top: 50px;
    color: #ffffff;
      display: flex;
     justify-content: center;
     align-items: center;
     height: 600px;
     flex-direction: column;
}`

styles/style.css
```diff
@@ -6,3 +6,178 @@ Green: #00B172
...
+    align-content: center;
+  }
+
+  .blok{
+      font-size: 1.5em;
+      text-align: center;
+      box-sizing: border-box;
+      padding: 20px;
+      margin: 50 5% 0 5%;
+      width: 40%;
```
LloydC, line 133: Since you have three articles, it would flow better if you changed line 133 to `width:33%;`

styles/style.css
```diff
@@ -6,3 +6,178 @@ Green: #00B172
...
+      flex-direction: column;
+      width: 33%;
+}
+
+.blokje{
+    text-align: left;
+    box-sizing: border-box;
+    font-size: 1em;
+}
+
```
LloydC, line 166: Line 167-169 should be removed

index.html
```diff
@@ -6,13 +6,85 @@
...
+<article class="blok">
+  <img src = "images/devices-icon.png" alt="Devices"/>
+  <h3>Stream Everywhere</h3>
+  <p class="text">Stream music on your smartphone, tablet or computer</p>
+
+</article>
+</div>
+
+</section>
+
```
LloydC, line 61: It would be easier if you used `Icon white` as a background image for `<section id="green"`. Then you would use flexbox on `<div class= "drierij">` to get `<div class="textrij">` and `Spotify app` to appear side by side better

PR #2106, opened 2020-10-14
[FT-AMS-102020] Emad Ghnim

Emad-doz:


index.html
```diff
@@ -5,14 +5,86 @@
...
+          <ul>
+            <li><a href="#">Premium</a></li>
+            <li><a href="#">Discover</a></li>
+            <li><a href="#">Help</a></li>
+            <li><a href="#">Download</a></li>
+          </ul>
+        </div>
+      </nav>
+    
+
```
wax-bear, line 31: Are these blank lines necessary?

index.html
```diff
@@ -5,14 +5,86 @@
...
+        <p>Enjoy new music every Monday with your own personal playlist. Or sit back and enjoy Radio.</p>
+
+      </section>
+
+      <section>
+        <img src="./images/spotify-app.jpg" alt="spotify-app">
+      </section>
+    </footer>
+
+
```
wax-bear, line 86: Same question here regarding the blank lines.

styles/style.css
```diff
@@ -6,3 +6,128 @@ Green: #00B172
...
+    text-align: center;
+    margin-top: 50px;
+}
+
+#general-information hr {
+    border: 1px solid #00B172;
+    width: 200px;
+}
+
+.general-information-container {
```
wax-bear, line 79: Is there a reason why you use a class here rather than an id? If not, as you are not using this styling anywhere else, I would use an id. 
And could you nest this in the #general-information id somehow?

wax-bear (commented):
Great stuff Emad!

Your clone starts to look like the original. Would it be possible to identify the remaining differences with the original and iterate towards a definite clone version?

Please check out my comments regarding CSS and HTML.

Please reach out to me if you have any question.

PR #2109, opened 2020-10-14
[AMS-FT-012020]delia.drel

deframed:
I use padding and margin quite a lot to align more precisely. Probably it’s not the right way. So, what can you suggest?

Adding margin & padding has makes the box bigger, which has an effect on on the container where the box is in. I know there’s a way to include merging/padding/border in the box but then you need to adjust the content of the box. What is the right way to do?

I was also a bit confused with the float property. For me it made more sense to put this property in ul. But then it took the height of the header. Dimensions stayed the same when I moved the float property to il. Don’t really get this one.

### DIFF
--- a/index.html
+++ b/index.html
@@ -6,13 +6,73 @@
     <meta http-equiv="X-UA-Compatible" content="ie=edge" />
     <title>Spotify Clone</title>
     <!-- don't forget to link your styles -->
+    <link rel="stylesheet" href="./styles/style.css">
+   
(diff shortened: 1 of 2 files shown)


wax-bear (commented):
Hey Delia,

There is still some margin/padding missing around/in the green panel. What could be interesting, if you feel you are overdoing the margins, is to inspect your elements and then play around (enable/disable) with the style properties. That can tell you whether all the CSS is necessary or not.

Overall your clone looks good!

Please check out my comments for HTML/CSS improvements.

PR #2113, opened 2020-10-15
[FTAMS102020]issaGruber

GruberI:


styles/style.css
```diff
@@ -1,8 +1,160 @@
...
+.header-text > p {
+    width: 45%;
+    margin-left: 360px;
+    text-align: center;
+    font-size: 22px;
+    font-weight: 100;
+}
+
+section {
+     padding-bottom: 30px;
```
wax-bear, line 63: The indentation seems off here.

styles/style.css
```diff
@@ -1,8 +1,160 @@
...
+
+section > h2 {
+    display: flex;
+    justify-content: center;
+    text-align: center;
+    font-size: 30px;
+    margin-top: 50px;
+}
+
+#section-div {
```
wax-bear, line 74: Because you are using an id, perhaps be more specific in the naming of it. What section is referred to?

wax-bear (commented):
Opening the project in the browser, the page looks really good.

Two minor details could be enhanced. In the example website: 1. there is no margin left and right of the "concert" image, 2. there is more white space around the green section.

Your HTML and CSS look well structured.

GruberI:
Hi @waxbear thank you for your suggestions. I put in both font links
because I was having trouble with font-weight being overridden through out
my code when I tried to make <headers> bolder or <p> thinner. Is there a
reason that happens?

Also, the styling wouldn't stay consistent if I made my screen larger or
smaller. I think that's called being responsive? Is there a way to help
with that or do I need to wait till we get to grids?

Thank you again!

wax-bear:
Hey Issa,

I see two problems related to the imported font. Firstly, two link tags are used. This is not necessarily bad, but if they are referencing the same resource I would use one link tag and then process all "wght" values in one URL rather than two.
Secondly, the font weights imported (400 and 600) do not correspond to the font weight specified in the CSS (100). I would fix the latter issue to see whether the problem you describe persists.

Indeed, CSS can be used to make the page render in a specified way with different screen sizes/devices. This can be done by using media queries, as described in the following link: https://www.w3schools.com/css/css_rwd_mediaqueries.asp.

Happy coding!

Thomas

PR #2117, opened 2020-10-15
[FTAMS102020] Sabela Pena Garcia  

defSabe:
Iterations 1 and 2, of 4.

### DIFF
--- a/index.html
+++ b/index.html
@@ -2,12 +2,38 @@
 <html lang="en">
   <head>
     <meta charset="UTF-8" />
+    <link href="https://fonts.googleapis.com/css2?family=Noto+Sans+KR:wght@300&display=swap" rel="stylesheet">
+    <link rel="stylesheet" href="styles/style.css">
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
     <meta http-equiv="X-UA-Compatible" content="ie=edge" />
     <title>Spotify Clone</title>
     <!-- don't forget to link your styles -->
   </head>
   <body>
+  <header>
+    <div>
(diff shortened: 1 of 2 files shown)


index.html
```diff
@@ -2,12 +2,38 @@
...
+
+
+    <section class="sectionOne">
+      <div>
+        <h1> Music for everyone. </h1>
+        <br>
+        <h3 class= "free">Spotify is now free on mobile, tablet and computer. <br> 
+          Listen to the right music, wherever you are.</h3>
+      </div>
+        </section>>
```
reMRKableDev, line 36: Looks like there is an extra ```>``` closing tag that snuck along here 🤭. 
You can remove it 😄 
```suggestion
        </section>
```

styles/style.css
```diff
@@ -6,3 +6,55 @@ Green: #00B172
...
+    font-family: 'Noto Sans KR', sans-serif;
+}
+
+.logo {
+    width: 17%;
+    margin-right: 1rem;
+    margin-bottom: 11px;
+}
+
+.sectionOne {
```
reMRKableDev, line 41: The ```width``` and ```height``` properties in this declaration can be omitted. This way the image stays full across the screen

```suggestion
.sectionOne {  background: url("../images/landing.jpg");
  background-size: cover;
  text-align: center;
  font-size: 40px;
  padding: 170px;
  font-family: "Noto Sans KR", sans-serif;
  color: azure;
  left: 0;
  }
```

reMRKableDev (changes requested):
Nice attempt @defSabe 🚀 

There are a few comments, I have left that you can check out and revise 👍🏾

PR #2126, opened 2020-10-16
[FTAMS102020] AlejandroPliakasBenitez

apliakas:


index.html
```diff
@@ -1,18 +1,92 @@
...
+    <link rel="stylesheet" href="./styles/style.css">
   </head>
   <body>
-    Premium Discover Help Download Music for everyone. Spotify is now free on mobile, tablet and computer. Listen to the
-    right music, wherever you are. What’s on Spotify? Millions of Songs There are millions of songs on Spotify HD Music
-    Listen to music as if you were listening live Stream Everywhere Stream music on your smartphone, tablet or computer
-    It’s as yeezy as Kanye West. Search Know what you want to listen to? Just search and hit play. Browse Check out the
-    latest charts, brand new releases and great playlists for right now. Discover Enjoy new music every Monday with your
-    own personal playlist. Or sit back and enjoy Radio.
+    <div class="container">
```
reMRKableDev, line 13: I like that you are using this element to create some spacing on your page. A suggestion for next time is that you could give the same CSS properties you've given to ```.container``` to the ```body``` and you would achieve the same result. Additionally, you wouldn't have to have this extra ```div``` on for your page. 👍🏾 

So your CSS file could instead have this:
```css
body {
  width: 1400px;
  margin: 0 auto;
}
```
instead of this:
```css
.container {
  width: 1400px;
  margin: 0 auto;
}
```

index.html
```diff
@@ -1,18 +1,92 @@
...
+  
+        <div class="boxes">
+          <a class="box">Premium</a>
+          <a class="box">Discover</a>
+          <a class="box">Help</a>
+          <a class="box">Download</a>
+        </div>
+      </nav>
+  
+      <section class="bigpicture">
```
reMRKableDev, line 27: Good to see you using semantic ```section``` HTML tag. I encourage you to use them more for the different sections you create on your pages 👍🏾

index.html
```diff
@@ -1,18 +1,92 @@
...
+            <p>Enjoy new music every Monday with your own personal playlist. Or sit back and enjoy Radio.</p>
+          </article>
+        </div>
+  
+        <div class="spotifylogo">
+          <img src="images/spotify-icon-white.png" width="150px">
+        </div>
+  
+        <div class="appimage">
+          <img src="images/spotify-app.jpg" height="650px">
```
reMRKableDev, line 86: A thought and suggestion here. When using the attributes ```width``` or ```height``` on an element, it isn't necessary to give ```px``` to the values. This is because the ```width``` and ```height``` attributes already specified in pixels natively by HTML.

Additionally, when using `````` tags, it is important to also include the ```alt``` attribute. This is because in case the image isn't able to load, the ```alt``` value will be displayed to the user. The ```alt``` value is just a string that is needed. 

So in the case of all the `````` tags, you can remove the "px" and include the ```alt``` attribute. See the suggestion below:

```suggestion
          <img src="images/spotify-app.jpg" height="650" alt="spotify-app-image">
```

styles/style.css
```diff
@@ -6,3 +6,123 @@ Green: #00B172
 White: #FFF
 
 */
+html {
```
reMRKableDev, line 9: Nice work with the CSS 😄

reMRKableDev (changes requested):
Good job on submitting this @apliakas 🚀 

Most of the work done is proper, however, I have left a few comments for you to check out and revisit 😄

PR #2127, opened 2020-10-16
[FTAMS102020] Alex Kantsadis

teof95:
I added also hover on nav buttons

### DIFF
--- a/index.html
+++ b/index.html
@@ -1,18 +1,90 @@
 <!DOCTYPE html>
 <html lang="en">
-  <head>
+
+<head>
     <meta charset="UTF-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
     <meta http-equiv="X-UA-Compatible" content="ie=edge" />
     <title>Spotify Clone</title>
+    <link rel="stylesheet" href="/styles/style.css">
+
     <!-- don't forget to link your styles -->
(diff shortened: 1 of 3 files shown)


styles/style.css
```diff
@@ -6,3 +6,155 @@ Green: #00B172
...
+    font-family: Arial, Helvetica, sans-serif;
+    font-size: 14px;
+    margin: 0;
+    height: 100%;
+}
+
+body {
+    background-size: cover;
+    height: 100%;
+    width: 1000px;
```
LloydC, line 21: Your website would be better if you removed `margin: 0 auto;` and changed line 21 to `width: auto;`

index.html
```diff
@@ -1,18 +1,90 @@
...
+
+    <h2 class="top-title">What’s on Spotify?</h2>
+
+
+    <div class="firstSection">
+
+        <div class="box">
+            <img class="smallIcons" src="images/music-icon.png" alt="smallIcons" />
+            <h3 class="psecond">Millions of Songs</h3>
+            <p id="smallpar"> There are millions of songs on Spotify</p>
```
LloydC, line 43: It would be more correct to replace `id="smallpar"` with `class="smallpar"` because you're using it again on line 50 and 57. An id is supposed to be used only once

PR #2407, opened 2021-03-24
[FTAMS032021] SilviaRodriguez - Jamal, Andrea

silrodcar:


styles/spotify-clone.css
```diff
@@ -0,0 +1,168 @@
...
+    display: flex;
+    flex-direction: column;
+    justify-content: space-around;
+    align-items: flex-start;
+
+}
+.green-box-p {
+    padding-top: 20px;
+    font-size: 1.2em;
+    font-weight: 200px;
```
reMRKableDev, line 160: Using `px` on `font-weight` isn't considered the correct value type. See here for the correct types of values to use --> https://cssreference.io/property/font-weight/

You can remove the `px`

index.html
```diff
@@ -5,14 +5,78 @@
...
+        <ul class="nav-links">
+          <li> <a href="#">Premium</a></li>
+          <li> <a href="#">Discover</a></li>
+          <li> <a href="#">Help</a></li>
+          <li> <a href="#">Download</a></li>
+        </ul>
+      </nav>
+
+    </header>
+     <section class= "box">
```
reMRKableDev, line 24: Using `<main>` as a parent element for all these sections would be good practice

reMRKableDev (commented):
Nice work @silrodcar 🚀 

Good to see you using some semantic tags in your HTML. Also, the styling looks great. I see you've used some flexbox properties which is always appreciated. All in all, good stuff. 

I've left some comments that you can check out

PR #2450, opened 2021-04-13
[PAR-FT-042021] Jordan Caley

jordanacaley:
To do the landing banner I got some ideas from this link (https://css-tricks.com/text-blocks-over-image/) but I am curious to see other approaches. 

I think my major issues are just with responsiveness since I didn't add media queries.

Thanks!

### DIFF
--- a/index.html
+++ b/index.html
@@ -53,10 +53,22 @@
       </div>
     </section>
 
-      
+    <section>
+      <div class="green">
+        <div class="flex-column left">
+          <h2>It’s as yeezy as Kanye West.</h2>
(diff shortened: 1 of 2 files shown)


Gorlanova:
Nice trick for the image ! I personally put it as a background image. To me, every image that has no semantic meaning (backgrounds, decorative stuff) will be dealt with CSS as much as possible, in order to leave html elements for meaningful images (icons, logos as link content, etc...) so screen readers are not overwhelmed with purely decorative stuff, but hey, the result is good ! Nice use of relative / absolute positioning :)

PR #2452, opened 2021-04-13
[PAR-WDFT-042021] Alice LEFEVRE

aliceislucy:
Hello :)

I struggled on the " top : 12.5vh " property which I tried to set on my 2nd section. It didn't work because I forgot the position was in static, you helped me for that in zoom :) and that was the very main struggle I had, the rest was rather accessible for me, I'm starting to get the hang of it !

I was wondering what I could do about the last picture (the spotify-app image) because it looks pretty okay when the browser window is at a normal size but when I play around with it and reduce the sides of the browser window to the extreme, the image gets out of the section and goes out of the visible part of the window. I wish it would go under the text or anything that looks a little less awkward ! :laughing: 

Thanks for everything :)

### DIFF
--- a/index.html
+++ b/index.html
@@ -1,18 +1,120 @@
 <!DOCTYPE html>
+
 <html lang="en">
   <head>
     <meta charset="UTF-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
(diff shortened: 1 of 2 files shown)


Gorlanova:
Hey Alice ! Flex-Wrap is what you want to check ;)

PR #2483, opened 2021-05-27
[WebDev-BER-PT-052021] Julia Häußler

juliahaeussler:
done with header, 
working on the first section now, continuing friday morning


styles/style.css
```diff
@@ -6,3 +6,61 @@ Green: #00B172
...
+    font-size: 1.25em;
+}
+
+
+.topics {
+    display: flex;
+    flex-direction: row;
+    justify-content: space-around;
+    padding: 0 50px;
+}
```
numbing, line 66: so far its really good, also if you have time try to finish it but if you dont have time dont worry we going to touch css again

PR #2485, opened 2021-05-27
[WD-PT-BER-MAY2021] Franziska Pendzialek

franziskapendzialek:


styles/style.css
```diff
@@ -6,3 +6,119 @@ Green: #00B172
...
+.middle-container img {
+    width: 100px;
+}
+
+.middle-container h3 {
+    color: #00B172;
+    padding: 2%;
+}
+
+/* Feature List */
```
numbing, line 124: so far its really good, also if you have time try to finish it but if you dont have time dont worry we going to touch css again

franziskapendzialek: Thanks! 

Finished it this morning.

PR #2583, opened 2021-07-24
[WD PTRM] Daniel ten Damme

Danieltda:


IH-WebDev-TA-Remote:
Nice work Daniel! Almost pixel perfect! :eyes: Here's some feedback to further improve your solution if you have the time:
-> In your `.songs` class, you can apply a `max-width` property so that those flex-children have the same size. The outcome will be that they will be centered horizontally;
-> As for the white spotify icon in the last section, one way to center it is to make it a direct child of the div where you apply the green background, and using of the position relative (in the parent) and absolute (in the icon) so you can center it in the middle both vertically & horizontally :smile: 
Other than that, great solution Daniel! Keep up the good work :1st_place_medal:

PR #2584, opened 2021-07-26
[WD PTRM]Cheryl Huang

CherylHuang1988:


IH-WebDev-TA-Remote:
Excellent job Cheryl! Almost pixel perfect! :tada: As for the white spotify icon in the last section, one way to center it in the green section is to make it a direct child of the div where you apply the green background, and using of the position relative (in the parent) and absolute (in the icon) so you can center it in the middle both vertically & horizontally :smile:
Other than that, amazing job once again! :1st_place_medal:

PR #2587, opened 2021-07-26
[WD-PT-Remote - juli 2021] Juan Carlos Eugenio

juaneugenio:


IH-WebDev-TA-Remote:
Fantastic solution to this lab Juan! Bonus points for figuring out how to center the white spotify icon in the last section! :clap: 
Here's some feedback in case you want to further improve your solution in the future:
-> In the navbar, you can apply a `max-width` property to your class `navbar-list`, giving it a specific value (ex: 350px). The outcome will be that the links will not span further than that amount, but can shrink if needed be (for smaller screens);

Other than that, amazing job! Keep it up! :1st_place_medal:

PR #2637, opened 2021-09-23
WDPT-2021-REMOTE-KARLA-JARAMILLO

karlajaramillo:


karlajaramillo:
Hi Tomas!
I think I tried to do my best! And I'm happy with the result, but I know I'm still in the process of learning. I tested the page on a number of devices, using dev tools and it seems to work well. I changed flex properties to make it responsive and add hover transition for the navigation list.

Thanks!!!

PR #2733, opened 2021-11-19
[ WDPT SAO 11.2021 ] Lucas Mendes Gabriel

lucas-emg:
Complete clone of the Spotify page:
For Iteration 2 I have used flex box and I have given each of the three items fixed widths and used Space-Evenly for the Justify Content- which made so that each div had the exact same size and distance from each other.

For the Iteration 3, instead of positioning the logo using position absolute, I have applied it as a background image and positioned it on the center of page, which made it so that it was not on the flow of the page and gave me more space to align the text div and the image div accordingly, without conflicting with the image. :)


AnaCRDEL:
Muito bem! Ótimas soluções! 👏

PR #2865, opened 2022-04-15
[FTLIS042022] Francisco Bento

FranciscoGBento:


styles/style.css
```diff
@@ -6,3 +6,127 @@ Green: #00B172
...
+
+* {
+  font-family: sans-serif;
+}
+
+nav {
+  position: fixed;
+  display: flex;
+  justify-content: space-between;
+  background-color: #fff;
```
webLisbonTA, line 18: you could also add `width: 100%` so you would remove the right space on the navbar after adding position:fixed.

styles/style.css
```diff
@@ -6,3 +6,127 @@ Green: #00B172
...
+}
+
+.spotifylogo {
+  width: 150px;
+  height: 50px;
+  display: flex;
```
webLisbonTA, line 24: spotifylogo is a child element of your nav flex container, so you don't need to add display: flex and justify-content -- these are properties only available to the parent container.

webLisbonTA (commented):
Well done, Francisco!

PR #2868, opened 2022-04-15
FTLIS042022 Anna Egger

annagramcodes:
I couldn't  figure out how to properly align the card icons on the 2. section.  After setting the height of the icons to the same value, they still are positioned differently.


webLisbonTA:
Well done, Anna!
Very good code organization and class names.

I couldn't see your issue on the icons, if you want to review it, please let one of the TAs know.

PR #2870, opened 2022-04-15
cant link

EvgeniiaRiaguzova:


Enes-Ironhack:
You got confused with the naming of sections. (section2.1 is not correct)
You should look deeper into flexboxes ;)

PR #2879, opened 2022-04-16
CSS Spotify Clone

nataliabuzila:


Enes-Ironhack:
We can see that you ran out of time. But that was a great start ! keep working and you will eventually rock this project ;)

PR #2887, opened 2022-04-17
[FTLIS042022] Chloé Faurie

chloe4E:
well well well the app image is still not in the right place :/

### DIFF
--- a/index.html
+++ b/index.html
@@ -5,14 +5,80 @@
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
     <meta http-equiv="X-UA-Compatible" content="ie=edge" />
     <title>Spotify Clone</title>
+    <link rel="stylesheet" href="styles/style.css" />
     <!-- don't forget to link your styles -->
   </head>
   <body>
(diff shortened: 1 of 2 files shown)


index.html
```diff
@@ -5,14 +5,80 @@
...
+      <ul id="navbar">
+        <li>Premium</li>
+        <li>Discover</li>
+        <li>Help</li>
+        <li>Download</li>
+      </ul>
+    </nav>
+
+    <section id="large-image" class="center-me">
+      <div id="image-landing">
```
webLisbonTA, line 28: Best practice here is to add the image as a background of the section using css

styles/style.css
```diff
@@ -6,3 +6,115 @@ Green: #00B172
...
+  padding-left: 10%;
+  padding-right: 10%;
+  color: #fff;
+}
+
+#example h1 {
+  text-decoration: underline;
+}
+
+#app-image {
```
webLisbonTA, line 114: You were not able to adjust the img position because your #app-image selector refers to the `<section>` tag and not to the img itself. Try selecting the img and adjusting the css properties (eg. height) instead

webLisbonTA (commented):
Very good Chloé!

PR #2893, opened 2022-04-17
[FTAPR2022] João Morgado

JoaoMorgado8:


styles/style.css
```diff
@@ -6,3 +6,166 @@ Green: #00B172
...
+  align-content: stretch;
+}
+.features hr {
+  height: 7px;
+  width: 670px;
+  background-color: #00b172;
+  border: none;
+}
+
+.articles {
```
webLisbonTA, line 97: to make the 3 articles stay in a single row, you can adjust the width of each article card and also the width of the icon image

webLisbonTA (commented):
Well done, João!

PR #2894, opened 2022-04-17
[FTAPR2022] Ana Pinho

AnaGonPinho:


index.html
```diff
@@ -5,14 +5,80 @@
...
+     <li><a href="#">Discover</a></a>   
+     <li><a href="#">Help</a></li> 
+     <li><a href="#">Download</a></li> 
+      </ul>
+    
+    </div>
+   
+  </header>
+
   <body>
```
webLisbonTA, line 27: the body tag should come after the <head> and wrap all your content.
the <header>  is related with the content, so it should also come inside the <body>

index.html
```diff
@@ -5,14 +5,80 @@
...
     <title>Spotify Clone</title>
-    <!-- don't forget to link your styles -->
+    <link rel="stylesheet" href="./styles/style.css" />
   </head>
+  <header>
+  <div class = "logotipo">
+    <img src="./images/spotify-logo.png" alt="spotify-logo"/>
+  </div>
+  
+  <div class="nav-bar">
```
webLisbonTA, line 15: You could also use the semantic <nav>. tag here

index.html
```diff
@@ -5,14 +5,80 @@
...
   <body>
-    Premium Discover Help Download Music for everyone. Spotify is now free on mobile, tablet and computer. Listen to the
-    right music, wherever you are. What’s on Spotify? Millions of Songs There are millions of songs on Spotify HD Music
-    Listen to music as if you were listening live Stream Everywhere Stream music on your smartphone, tablet or computer
-    It’s as yeezy as Kanye West. Search Know what you want to listen to? Just search and hit play. Browse Check out the
-    latest charts, brand new releases and great playlists for right now. Discover Enjoy new music every Monday with your
-    own personal playlist. Or sit back and enjoy Radio.
+    <section id="music4every1" class = "subtitle">
+     <h1> Music for everyone.</h1>
+     <div> <p>Spotify is now free on
```
webLisbonTA, line 30: Usually we avoid using <br>, because of accessibility. If you need to adjust the line of the paragraph, try creating the second line in a new <p> tag or adjusting the width of the paragraph container

Reference: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/br

styles/style.css
```diff
@@ -6,3 +6,127 @@ Green: #00B172
...
+  object-fit: contain;
+}
+
+.nav-bar li {
+  display: inline-block;
+  margin: 20px;
+}
+
+#music4every1 {
+  min-height: 150% !important;
```
webLisbonTA, line 38: No need for !important here

webLisbonTA (commented):
Well done, Ana!

PR #2895, opened 2022-04-17
[FTWD0422] Catarina Salvado

catarinapsalvado:


styles/style.css
```diff
@@ -6,3 +6,161 @@ Green: #00B172
...
+    color: #1A1A1A;
+    background-color: #FFF;
+    margin: 0px;
+    padding: 0px;
+    font-size: 16px;
+    font-family:Verdana, Geneva, Tahoma, sans-serif
+}
+
+header {
+    display:inline-block;
```
webLisbonTA, line 20: don't need these properties, you are setting `display: flex` just below it

styles/style.css
```diff
@@ -6,3 +6,161 @@ Green: #00B172
...
+    flex-wrap: wrap;
+    justify-content: center;
+    
+}
+
+.section1 > h1{
+    color: #FFF;
+    font-size: 80px;
+    font-weight: 500;
+    display: flex;
```
webLisbonTA, line 73: no need for display: flex here.
h1 is a child element of your section1, display: flex is a property that goes to the parent container only.

webLisbonTA (commented):
Very good, Catarina!

PR #2896, opened 2022-04-17
[FTLIS042022] João Gonçalves

joaopdg:
lab solved

### DIFF
--- a/styles/style.css
+++ b/styles/style.css
@@ -6,3 +6,145 @@ Green: #00B172
 White: #FFF
 
 */
+
+body {
+  font-family: Arial, Helvetica, sans-serif;
+  margin: 0;
+}
+
+header {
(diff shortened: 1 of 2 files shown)


styles/style.css
```diff
@@ -6,3 +6,145 @@ Green: #00B172
...
+  color: black;
+}
+
+.center-part > h2 {
+  margin-left: 540px;
+  margin-top: 40px;
+  margin-bottom: 15px;
+  border-bottom: 4px solid #00b172;
+  width: 220px;
+}
```
webLisbonTA, line 80: Try to centralize the h2 using other techniques, like display flex, etc...
using margins for big alignments like this is not a  good practice

webLisbonTA (commented):
Very good, João!

PR #2897, opened 2022-04-18
[FTLIS0422] Duarte Alves

drttyy:


styles/style.css
```diff
@@ -6,3 +6,134 @@ Green: #00B172
...
+  padding-right: 50px;
+  padding-top: 10px;
+}
+nav {
+  background-color: #fff;
+  height: 50px;
+}
+nav > img {
+  width: 15%;
+  flex: left;
```
webLisbonTA, line 32: i believe this property doesn't exist

styles/style.css
```diff
@@ -6,3 +6,134 @@ Green: #00B172
...
+.music-box img {
+  width: 35%;
+}
+.sec2 > div {
+  display: flex;
+  flex-direction: row;
+  flex-wrap: nowrap;
+  justify-content: space-around;
+  align-items: center;
+  align-content: center;
```
webLisbonTA, line 90: `align-content` just works on multiline flex container.
in this case align-items is already doing the job!

styles/style.css
```diff
@@ -6,3 +6,134 @@ Green: #00B172
...
+  position: relative;
+  border: 5px solid #00b172;
+}
+.music-box {
+  flex-direction: column;
+  flex-wrap: wrap;
+  box-sizing: border-box;
+  text-align: center;
+  font-size: 2.5em;
+  padding-left: 7%;
```
webLisbonTA, line 78: By adding padding-left only you are pulling your content to the right

webLisbonTA (commented):
Keep going, Duarte!
Try adding the .spotify-app and the .spotify-white on the last section. 
There is not html tag for these selectors.

PR #2940, opened 2022-06-01
done

ElvanIpeker:
seems not exactly the same, sorry🙈😊


DaniCalvente:
Great effort Elvan!!👍.  You need to work harder on the styles, but remember you can review the class when Andre worked on this lab, and use it as a guide. Keep pushing!!💪💪

PR #3271, opened 2023-10-11
[LIS-FT-OCT23] Mariana Carvalho

marianacarvalho-it:


webLisbonTA:
Hi Mariana, 
This lab is very good to practice the use of HTML and CSS. 
I would advise you to finish the lab until this weekend to practice.

